PR #37, opened 2025-05-13
Support resolving merge request discussion notes

ncrum:
Allow the ability to resolve/unresolve notes in merge request discussions through the MCP API. Basically, the Gitlab API expects either `body` or `resolved`, but not both. The existing code here required `body` to be provided. I simply made that field optional.

Here is a reference to where this is called out in the API docs: https://docs.gitlab.com/api/discussions/#modify-an-existing-merge-request-thread-note

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1027,7 +1027,7 @@ async function updateMergeRequestNote(
   mergeRequestIid: number,
   discussionId: string,
   noteId: number,
-  body: string,
+  body?: string,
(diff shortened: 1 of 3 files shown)


zereight:
I will check it out! Thx! 🥰

zereight:
https://github.com/zereight/gitlab-mcp/releases/tag/v1.0.37

PR #38, opened 2025-05-13
Adds threaded comment support for merge requests

ncrum:
Adds two new MCP tools to improve the comment experience on merge requests. These changes allow threads to be created with inline code suggestions and notes to be added to existing threads.

1. `create_merge_request_thread` - Start new threads on merge requests, with support for diff notes and comment positioning
2. `create_merge_request_note` - Reply to existing threads

`create_note` as is does not support this functionality, so figured I would extend to support these two new tools instead.


zereight:
Wow..! thank you. It look so great.

zereight:
https://github.com/zereight/gitlab-mcp/releases/tag/v1.0.37

PR #47, opened 2025-05-22
feat: add support for creating and updating issue notes

svengt:
- Added create_issue_note to add a new note to an existing issue thread
- Added update_issue_note to modify an existing issue thread note
- Similar to existing merge request note functions but for issues


zereight:
thx!

PR #65, opened 2025-05-29
FEAT: ci push docker hub

iwakitakuma33:
@zereight If you need auto pushing the image on release, you can use ci workflow.

below is local simulation. 
`act release -W .github/workflows/docker-publish.yml --secret-file .secrets -e event.json`

To implement locally, you need to enter the credentials in .secrets.

When using ci, you also need to register the secret in GitHub.

If you manage it on your Docker Hub, you will also need to modify the readme file.


zereight:
👍 👍 👍 👍 👍 👍 👍 👍 👍

PR #78, opened 2025-06-02
FIX: issue param

iwakitakuma33:
https://docs.gitlab.com/api/issues/#:~:text=Return%20issues%20for%20the%20given%20scope%3A%20created_by_me%2C%20assigned_to_me%20or%20all.%20Defaults%20to%20created_by_me


iwakitakuma33:
@bbang-dduck thanks

PR #100, opened 2025-06-12
feat: add cookie-based authentication support for GitLab instances 🍪

mattweg:
- Add GITLAB_AUTH_COOKIE_PATH environment variable support
- Handle #HttpOnly_ prefix in cookie files properly
- Enable redirect following when cookies are present
- Maintain compatibility with existing token-based auth

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,8 @@ $ sh scripts/image_push.sh docker_user_name
 - `USE_GITLAB_WIKI`: When set to 'true', enables the wiki-related tools (list_wiki_pages, get_wiki_page, create_wiki_page, update_wiki_page, delete_wiki_page). By default, wiki features are disabled.
 - `USE_MILESTONE`: When set to 'true', enables the milestone-related tools (list_milestones, get_milestone, create_milestone, edit_milestone, delete_milestone, get_milestone_issue, get_milestone_merge_requests, promote_milestone, get_milestone_burndown_events). By default, milestone features are disabled.
 - `USE_PIPELINE`: When set to 'true', enables the pipeline-related tools (list_pipelines, get_pipeline, list_pipeline_jobs, get_pipeline_job, get_pipeline_job_output, create_pipeline, retry_pipeline, cancel_pipeline). By default, pipeline features are disabled.
+- `GITLAB_AUTH_COOKIE_PATH`: Path to an authentication cookie file for GitLab instances that require cookie-based authentication. When provided, the cookie will be included in all GitLab API requests.
+- `GITLAB_AUTH_COOKIE_PATH`: Path to an authentication cookie file for GitLab instances that require cookie-based authentication. When provided, the cookie will be included in all GitLab API requests.
 
 ## Tools 🛠️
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -111,6 +111,8 @@ $ sh scripts/image_push.sh docker_user_name
 - `USE_GITLAB_WIKI`: When set to 'true', enables the wiki-related tools (list_wiki_pages, get_wiki_page, create_wiki_page, update_wiki_page, delete_wiki_page). By default, wiki features are disabled.
 - `USE_MILESTONE`: When set to 'true', enables the milestone-related tools (list_milestones, get_milestone, create_milestone, edit_milestone, delete_milestone, get_milestone_issue, get_milestone_merge_requests, promote_milestone, get_milestone_burndown_events). By default, milestone features are disabled.
 - `USE_PIPELINE`: When set to 'true', enables the pipeline-related tools (list_pipelines, get_pipeline, list_pipeline_jobs, get_pipeline_job, get_pipeline_job_output, create_pipeline, retry_pipeline, cancel_pipeline). By default, pipeline features are disabled.
+- `GITLAB_AUTH_COOKIE_PATH`: Path to an authentication cookie file for GitLab instances that require cookie-based authentication. When provided, the cookie will be included in all GitLab API requests.
+- `GITLAB_AUTH_COOKIE_PATH`: Path to an authentication cookie file for GitLab instances that require cookie-based authentication. When provided, the cookie will be included in all GitLab API requests.
```
zereight: Can you remove the duplicate lines?

zereight (changes requested):

README.md
mattweg: Done.

index.ts
```diff
@@ -267,6 +328,11 @@ const DEFAULT_FETCH_CONFIG = {
   },
 };
 
+// Only add redirect following if we have a cookie
+if (authCookie) {
+  DEFAULT_FETCH_CONFIG.redirect = 'follow';
+}
+
```
zereight: It's neat code. However, I am concerned that it seems to be becoming a code that harms the immutability of constants.

I think we need to refactor it for clean code quality as a whole.

If you have a better direction, please suggest it. If not, we will distribute it after the end of the work.

Thank you

mattweg: Great feedback. I'll re-approach it with immutability in mind.

iwakitakuma33: @mattweg thanks so much for your enthusiastic improvement.

mattweg: I was thinking about how to solve this immutably. 

### Option 1

```javascript
// Create fetch config based on authentication method
const createFetchConfig = (authCookie) => {
 return {
   ...DEFAULT_FETCH_CONFIG,
   // Only add redirect following if we have a cookie
   ...(authCookie ? { redirect: 'follow' } : {})
 };
};

// Then use this function when creating the fetch configuration
const fetchConfig = createFetchConfig(authCookie);
```

### Option 2
```javascript
const DEFAULT_FETCH_CONFIG = {
 headers: {},
 redirect: 'follow'  // Always follow redirects
};
```

I don't have a strong preference. The follow setting won't really hurt anything, but it's not needed for anything other than this type of auth.

PR #101, opened 2025-06-14
feat: add cookie-based authentication support for enterprise GitLab 

mattweg:
Add support for Netscape cookie file authentication to enable access to enterprise GitLab instances that use SSO/OAuth2 redirects.

- Add GITLAB_AUTH_COOKIE_PATH environment variable
- Implement cookie jar with proper domain handling for redirects
- Use conditional fetch assignment: cookie-enabled when path configured
- Maintains backward compatibility: no cookies = original behavior
- Zero changes to existing fetch() calls throughout codebase

Enables authentication flows like: curl -L -b ~/.midway/cookie Useful for enterprise environments with federated authentication.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## [1.0.62] - 2025-06-10
+
+### Fixed
+
+- 🔐 **Private Token Authentication Fix**: Fixed Private-Token header authentication for GitLab API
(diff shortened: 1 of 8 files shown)


mattweg (commented):
I'm back again. After testing across several platforms, I realized the previous solution (#100) won't work for certain cookie formats. This is the cleanest solution I found.

PR #130, opened 2025-07-02
[draft] User based authentication methods

elee1766:
This PR completely refactors how gitlab authentication is done (along with configuration), removing a large amount of global variables.

this is required in order to support passing auth from the client to the server, instead of authorizing only at the server level.

the PR introduces two authentication schemes:

1. pat-passthrough, which allows clients to send `gitlab-token` header with a PAT that will be used for all their requests to the MCP

2. (experimental) oauth2, which uses a really hacky oauth2 proxy which fakes dynamic client registration in order to satisfy the MCP authorization standard, as gitlab themselves, or at least, my self hosted instances, does not seem to support dynamic client registration.

it is still a draft because the oauth2 backend is not well refined yet, has no real security review/audit, and has only an in-memory backend.

the advantage of this is that this allows the mcp server to be run and hosted at a single url by either the gitlab server operator or a trusted party, and developers no longer need to install the mcp server, and dont configure as many variables on their local clients.


elee1766:
hi @bbang-dduck 

i understand there are a lot of big changes here, but some of them were required in order to support the features,and others are good changes that helped with the development of this feature, and also are i think good for the quality of the code base in general. I am happy to change things and structures if you dont like certain things but at the very least the refactoring and splitting up of the index.ts felt basically required given how complicated this feature was to implement.

i have tested this locally using our companies gitlab instance so far.
 
 if @AndrewFarley you are interested in helping test this (both modes) that would be helpful.

elee1766:
hi @zereight @iwakitakuma33 

is there anything that you would like me to do / change that would accept this being merged?

iwakitakuma33:
@elee1766 thanks.

I am creating a test environment that can be developed locally.
I will try it out there.
https://github.com/zereight/gitlab-mcp/pull/171

qvalentin:
Hi @elee1766 thanks implementing this. Currently I can not build your branch but I would be interested in testing.

qvalentin:
This is my .env:

```
# GitLab API Configuration

# Test Configuration (for integration tests)

# MCP Transport Mode (Optional)
# Description: 
# When multiple transport modes are enabled, the server will use the following priority:
# 1. **Streamable HTTP** (if `STREAMABLE_HTTP=true`) - Highest priority
# 2. **SSE** (if `SSE=true` and `STREAMABLE_HTTP!=true`) - Medium priority  
# 3. **Stdio** (if `SSE!=true` and `STREAMABLE_HTTP!=true`)
export SSE=false
export STREAMABLE_HTTP=true

# MCP Server Host With SSE Transport and Streamable Http Transport
export HOST=127.0.0.1
export PORT=3000

# Proxy Configuration (optional)
export HTTP_PROXY=
export HTTPS_PROXY=
export NO_PROXY=localhost,127.0.0.1


# GitLab OAuth2 Application Credentials
export GITLAB_OAUTH2_CLIENT_ID=ID
export GITLAB_OAUTH2_CLIENT_SECRET=SECRET

# GitLab OAuth2 Endpoints
export GITLAB_OAUTH2_AUTHORIZATION_URL=https://gitlab.example.com/oauth/authorize
export GITLAB_OAUTH2_TOKEN_URL=https://gitlab.example.com/oauth/token
export GITLAB_OAUTH2_ISSUER_URL=https://gitlab.example.com
export GITLAB_OAUTH2_BASE_URL=http://localhost:3000  # Your MCP server URL

# OAuth2 Redirect Configuration
export GITLAB_OAUTH2_REDIRECT_URL=http://localhost:3000/callback

export LOG_LEVEL=debug
```

qvalentin:
1. & 2. Yes its self-hosted.
3. Its a user level application since Im not an admin.
4. I added logging to verifyAccessToken and saw that the token was expired. This should be handled gracefully.
5. Good point, why is this not in the example.env.

Its working now :D. I just have to get the token with the modelcontextprotocol/inspector manual mode and then i can past it in my python project.
Thank you, I might add some comments to the code later.

docs/oauth2_proxy.md
```diff
@@ -0,0 +1,147 @@
...
+   - **Application ID**: Use this for `GITLAB_OAUTH2_CLIENT_ID`
+   - **Secret**: Use this for `GITLAB_OAUTH2_CLIENT_SECRET`
+
+## Configuration Examples
+
+### Development Setup (localhost)
+
+```bash
+# .env file
+GITLAB_API_URL=https://gitlab.com
```
qvalentin, line 65: I guess this should include the path: https://gitlab.com/api/v4

docs/oauth2_proxy.md
```diff
@@ -0,0 +1,147 @@
...
+GITLAB_OAUTH2_ISSUER_URL=https://gitlab.com
+GITLAB_OAUTH2_BASE_URL=http://localhost:3000
+GITLAB_OAUTH2_REDIRECT_URL=http://localhost:3000/callback
+```
+
+### Production Setup
+
+```bash
+# .env file
+GITLAB_API_URL=https://gitlab.company.com
```
qvalentin, line 80: should also include /api/v4

Edit: normalizeGitLabApiUrl already adds it

docs/oauth2_proxy.md
```diff
@@ -0,0 +1,147 @@
...
+GITLAB_OAUTH2_REDIRECT_URL=https://mcp.company.com/callback
+GITLAB_OAUTH2_DB_PATH=/var/lib/gitlab-mcp/oauth.db
+```
+
+## Database Storage
+
+By default, the OAuth2 proxy uses an in-memory SQLite database. For production use, specify a persistent database path:
+
+```bash
+GITLAB_OAUTH2_DB_PATH=/path/to/persistent/oauth.db
```
qvalentin, line 97: How would one do a production setup with multiple load-balanced instances of the MCP server?

package.json
```diff
@@ -32,8 +32,11 @@
     "format:check": "prettier --check \"**/*.{js,ts,json,md}\""
   },
   "dependencies": {
-    "@modelcontextprotocol/sdk": "^1.10.0",
+    "@modelcontextprotocol/sdk": "1.13.3",
+    "@node-rs/argon2": "^2.0.2",
```
qvalentin: Note that i had to run `npm i @node-rs/argon2-darwin-arm64` on macos.

index.ts
```diff
@@ -1,4523 +1,279 @@
...
-        const discussions = await listMergeRequestDiscussions(
-          project_id,
-          merge_request_iid,
-          options
-        );
-        return {
-          content: [{ type: "text", text: JSON.stringify(discussions, null, 2) }],
-        };
+    app.post("/messages",authMiddleware,  async (req: Request, res: Response) => {
+      const sessionId = req.query.sessionId as string;
```
qvalentin, line 95: We should validate that sessionId is an expected value. (string with min length of...?)

index.ts
```diff
@@ -1,4523 +1,279 @@
...
+            logger.error({e},"Transport error handling request:");
+            res.status(500).send("Internal server error");
+            return
+          }
+        }
+      } catch (error) {
+        logger.error('Streamable HTTP error:', error);
+        res.status(500).json({
+          error: 'Internal server error',
+          message: error instanceof Error ? error.message : 'Unknown error'
```
qvalentin, line 231: Never send unknown error messages back to the user.

index.ts
```diff
@@ -1,4523 +1,279 @@
...
-        };
-      }
+  if (streamableHttpEnabled) {
+    const transports : { [sessionId: string]: {
+      transport: StreamableHTTPServerTransport
+      tokenHash?: string
+    }} = {};
+    // Streamable HTTP endpoint - handles both session creation and message handling
+    app.post('/mcp', authMiddleware, async (req: Request, res: Response) => {
+      const sessionId = req.headers['mcp-session-id'] as string;
```
qvalentin, line 146: Should validate for UUID

index.ts
```diff
@@ -1,4523 +1,279 @@
...
+            await transport.handleRequest(req, res, req.body);
+          } catch (e) {
+            logger.error("Transport error handling request:", e);
+            res.status(500).send("Internal server error");
+            return
+          }
+        } else {
+          // Create new transport for new session
+          transport = new StreamableHTTPServerTransport({
+            sessionIdGenerator: () => randomUUID(),
```
qvalentin, line 184: Ok so we are generating the session id, so we should validate them later (see comment above)

index.ts
```diff
@@ -1,4523 +1,279 @@
...
+              // if we have a valid auth info here, either obtained from the passthrough token or oauth, we tie it to a session.
+              if(req.auth) {
+                transports[newSessionId].tokenHash = argon2.hashSync(req.auth.token, {
+                  salt: argon2Salt,
+                });
+                logger.debug({
+                  tokenHash: transports[newSessionId].tokenHash.slice(-8),
+                }, "auth session created for token")
+              }
+              logger.warn(`Streamable HTTP session initialized: ${newSessionId}`);
```
qvalentin, line 198: Info level?

index.ts
```diff
@@ -1,4523 +1,279 @@
...
-    }
-  });
+  }
 
   app.get("/health", (_: Request, res: Response) => {
     res.status(200).json({
       status: "healthy",
-      version: SERVER_VERSION,
-      transport: TransportMode.SSE
+      version: process.env.npm_package_version || "unknown",
```
qvalentin, line 241: I don't think we should provide an attack with the info what version we are running.

src/authentication.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Configure authentication middleware based on the configuration
+ * Supports OAuth2, PAT passthrough, and static PAT modes
+ */
+export async function configureAuthentication(app: Express): Promise<RequestHandler> {
+  // Default middleware that does nothing
+  let authMiddleware: RequestHandler = (_req: Request, _res: Response, next: NextFunction) => {
+    next();
+  };
```
qvalentin: This should not be used, since all possible branches set the value of authMiddleware. If it is still the default at the end we would have a configuration problem and should throw an error.

src/config.ts
```diff
@@ -0,0 +1,100 @@
...
+  HTTP_PROXY : process.env.HTTP_PROXY,
+  HTTPS_PROXY : process.env.HTTPS_PROXY,
+  NODE_TLS_REJECT_UNAUTHORIZED : process.env.NODE_TLS_REJECT_UNAUTHORIZED,
+  GITLAB_CA_CERT_PATH : process.env.GITLAB_CA_CERT_PATH,
+  // Use the normalizeGitLabApiUrl function to handle various URL formats
+  GITLAB_API_URL: normalizeGitLabApiUrl(process.env.GITLAB_API_URL || undefined),
+  GITLAB_PROJECT_ID: process.env.GITLAB_PROJECT_ID,
+
+
+  ARGON2_SALT: process.env.ARGON2_SALT || "change-me-in-production",
```
qvalentin: Maybe add the  "change-me-in-production" in example.env and not here.

src/authhelpers.ts
```diff
@@ -0,0 +1,50 @@
+import { config } from "./config.js";
+import { CookieJar, parse as parseCookie } from "tough-cookie";
+import fs from "fs";
+import path from "path";
+
+// Create cookie jar with clean Netscape file parsing
+export const createCookieJar = (): CookieJar | undefined=> {
```
qvalentin, line 7: This really confuses me, why were cookies even used in the first place?

src/oauth.ts
```diff
@@ -0,0 +1,505 @@
...
+import argon2 from '@node-rs/argon2';
+import { randomBytes } from 'crypto';
+
+// Custom provider that handles dynamic registration and maps to GitLab OAuth
+class GitLabProxyProvider extends ProxyOAuthServerProvider {
+  // Static async factory method
+  static async New(options: any): Promise<GitLabProxyProvider> {
+    // we put this here so we dont initialize this unless we are using the oauth provider
+    const Database = (await import('better-sqlite3')).default;
+    const db = new Database(config.GITLAB_OAUTH2_DB_PATH);
```
qvalentin, line 20: Persistency should really be decoupled from the OAuth implementation. This means:
1. Move all db logic to the a separate file
2. Create a clean interface that hides the persistency logic

We could then even have different implementations for this interface (in-memory, sqlite, redis?). Or we could consider using a ORM.

src/oauth.ts
```diff
@@ -0,0 +1,505 @@
...
+      token_hash: string;
+      client_id: string;
+      scopes: string;
+      expires_at: number | null;
+      timestamp: number;
+    }>;
+
+    // Find the matching token by verifying against each hash
+    let matchingRow = null;
+    for (const row of rows) {
```
qvalentin, line 265: Hmm this does not seem optimal, do we not have another way to identify the correct token? Like an id?

docs/oauth2_proxy.md
qvalentin: nvm, normalizeGitLabApiUrl adds it

src/oauth.ts
```diff
@@ -0,0 +1,505 @@
...
+import { AuthInfo } from '@modelcontextprotocol/sdk/server/auth/types.js';
+import { OAuthRegisteredClientsStore } from '@modelcontextprotocol/sdk/server/auth/clients.js';
+import { Request, Response } from 'express';
+import { logger } from './logger.js';
+import { Database } from 'better-sqlite3';
+import argon2 from '@node-rs/argon2';
+import { randomBytes } from 'crypto';
+
+// Custom provider that handles dynamic registration and maps to GitLab OAuth
+class GitLabProxyProvider extends ProxyOAuthServerProvider {
```
qvalentin, line 15: I read a little more about the spec and have one concern:

Do I understand correctly that the token that is created by the OAuth flow with GitLab is sent to the MCP client? The GitLabProxyProvider only validates that the token is stored in the db but then passes it on.

```js
gitlabToken: token, // Return the original token since we don't store gitlab_token anymore
```

So there is no token that is only between the MCP server and client but instead the GitLab token is used.

Gemini sees a problem with this and I'm sure its right:
```
The Critical Reasons for Issuing Your Own Token
1. Security: The Principle of Least Privilege
What the GitLab token can do: The token you get from GitLab might have permissions (scopes) to read the user's private repositories, post comments, or access their profile. This is called the "scope" of the token.
What the MCP client needs: The MCP client only needs permission to perform actions on your MCP server.
The Risk: If you gave the GitLab token to the MCP client, and that client was compromised, the attacker would have access to the user's entire GitLab account. By issuing your own token, you limit the potential damage. The compromised client can only access your service, not the user's GitLab.
```

I would also recommend looking for a established library to do the OAuth instead having so much custom code.

docs/oauth2_proxy.md
elee1766: out of scope here

src/authhelpers.ts
elee1766: i don't know, its in the original server and it said you had to read some sort of file for some cookie to access gitlab. no clue.

src/oauth.ts
elee1766: the persistence is specific to the oauth implementation.

different implementations for the interface is out of scope. could be done later for those who need it. 

this PR really only is showing that 

1. token passthrough is possible
2. DCR proxy over an oauth2 server that does not support DCR is possible

src/oauth.ts
elee1766: if you can find a library which supports https://datatracker.ietf.org/doc/html/rfc7591 on top of an existing oauth provider, then that would be preferred, but I could not find one

elee1766: > So there is no token that is only between the MCP server and client but instead the GitLab token is used.

the user holds their token in the client, the MCP simply makes sure that the token is a token that it issued through its system (since it remembers the argon2 hashes of tokens it has issued)

ideally, we would not need the oauth wrapper and gitlab would support dynamic client registration, auth tokens could be passed through without any internal database validation, but gitlab doesn't

src/oauth.ts
elee1766: we can't store a deterministic id because that would be saving an unsalted hash in the database, which I don't want to do.

if there was a non-deterministic id, then we would need to somehow append it to the api key, and then extract it on read, but if i did that, then the token would no longer be a valid gitlab token

src/oauth.ts
elee1766: > The Risk: If you gave the GitLab token to the MCP client, and that client was compromised, the attacker would have access to the user's entire GitLab account. 

correct. the token generated for use is a gitlab token, as if you were directly using the gitlab application. the mcp oauth layer exists in order to provide the facade of DCR, since it is required by the authorization spec, while still ultimately, just the gitlab application in the background.

if you compromised the server, you would receive keys. this is true.

this is an intended feature of the implementation. maintaining tokens for specifically the MCP seems like a bad idea. the MCP is not really its own oauth server, and delegates all authentication to gitlab via the configured application.

the gitlab MCP client needs a token that is scoped to YOUR user in order to send the request to gitlab as if YOU are the one doing the request. the role of the gitlab MCP is to impersonate the user - not act using a service account.

as a result - we store the salted hash of the token to validate that the token is one we issued (through the proxy), and we pass through the underlying token.

so gemini is right about what im doing, but otherwise, it is wrong. it does not understand that the mcp server must impersonate the underlying user. 

there is no way to run the mcp server in a way that you either 1. do not pass through the actual gitlab keys, or 2. do not store the gitlab keys on the server side.

it would be very not good to have a database with all the plaintext gitlab api keys of the users, so it's best security wise to not store the keys, and therefore make it so that the actual key sent to the client is an api key.

the only sensible option i can think of that doesn't change this is to use a gitlab administrator api key in the mcp server, and generate a temporary impersonation token for every user request. (you can't store them or else you have a database of keys again)

but the issue is that would require that the server have its completely own oauth implementation, and not just perform DCR and delegate to gitlab, which could be an even greater attack vector. the database is really only needed for the initial DCR, and as said above, we don't really even need to validate the tokens sent to us, as the token will be passed on to the request ultimately made to gitlab.

all in all - i think it makes most sense to do what is currently done, which is to pass through a valid gitlab token, but let me know if you have other thoughts or another solution.

if someone is capable of compromising your mcp.mygitlab.com, likely they could also compromise your mygitlab.com, which could steal your tokens anyways.

you should NOT use a gitlab mcp server hosted by somebody that you do not trust. that is a given.

index.ts
elee1766: there is no reason to validate for uuid.

index.ts
elee1766: there is no need to validate. string comparison is enough

package.json
elee1766: can maybe switch to an argon2 impl with less dependencies

index.ts
elee1766: there is no need to. waste of cycles.

index.ts
elee1766: security by obscurity.

src/config.ts
elee1766: i dont want the server to refuse to start up if the salt is not set. 

maybe a loud error message on startup if the salt is detected as the default salt

src/oauth.ts
elee1766: in truth, we don't really need to do this at all. we can also

1. just pass it through, because the token will fail authentication to the gitlab api anyways
2. just check the api token with a request against the gitlab api every time, since it's supposed to just be a valid gitlab token anyways

index.ts
elee1766: good catch.

src/oauth.ts
qvalentin: All valid points, I think the main decision that has to be made is to choose between the two risks:
- Giving the GitLab Token to the MCP client
- Storing the Gitlab Token in the MCP server


Regarding the variant where the server stores the Gitlab Tokens:
> but the issue is that would require that the server have its completely own oauth implementation, and not just perform DCR and delegate to gitlab

What about creating a second application in GitLab that has only the scope `openid` and using this for the authentication between the MCP client and the MCP server and using the application with `api` scope only for the authentication between the MCP server and GitLab?

qvalentin: Another option that avoids storing the tokens in plain text is to use Secret Sharing where the token is split in two parts and one part it given to the client the other part is stored in the server. Only with both parts you can then reconstruct the secret (you could use XOR to do this securely).

But I'm not an crypto expert and this is getting quite homebrew already.

elee1766: yeah, its going to have to be a little home brew, my hope was to do as little as possible, and try to delegate as much as I could to gitlab 

> What about creating a second application in GitLab that has only the scope openid and using this for the authentication between the MCP client and the MCP server and using the application with api scope only for the authentication between the MCP server and GitLab?


yeah, so you could potentially authenticate and store this "Identity" token, and then as I said, generate an impersonation token per request based off the user.

the issue is that this introduces a vector where a bug could allow users to execute actions as other users - a path that this current implementation makes difficult to accidentally do

zereight:
@elee1766 Can you fix conflict? I will test it.

elee1766:
@zereight hi, i've just merged

zereight:
@elee1766 Hi, plz resolve conflicts, I will approve it.

iwakitakuma33:
@zereight 
no problem!

elee1766:
@zereight @iwakitakuma33 just merged main

thank you both for the help getting this through. i hope that the PR will help future feature iteration as well.

zereight:
@iwakitakuma33 we will rollback this.

About. https://github.com/zereight/gitlab-mcp/issues/217

zereight:
This PR is rollback at 2.0.3 (same as 1.0.77).

I'm sorry. I will check about some issues.

PR #134, opened 2025-07-05
REMOVE TEST

iwakitakuma33:


package.json
```diff
@@ -23,7 +23,7 @@
     "deploy": "npm publish --access public",
     "generate-tools": "npx ts-node scripts/generate-tools-readme.ts",
     "changelog": "auto-changelog -p",
-    "test": "node test/validate-api.js && npm run test:server",
+    "test": "node test/validate-api.js",
```
zereight, line 26: I'm sorry for my negligence in managing these days.

I saw that there was a problem with the test code, but I wasn't doing anything about it.

I agree with this change, and let's write better test code together in the future.

zereight (approved):

package.json
iwakitakuma33: @zereight 
Since this test requires an actual GitLab environment, it would be better to check it locally first.

Issue #132, opened 2025-07-03
Json parse exception while init mcp server with version "version": "1.0.69",


lalchand92:
Stack trace

INFO  [2025-07-03 17:20:51,449] [metrics-metric-ingestion-reporter-1-thread-1] [MetricIngestionReporter]: Number of ServiceMetrics ingested - 947
INFO  [2025-07-03 17:20:51,965] [pool-21-thread-1] [StdioClientTransport]: STDERR Message received: Starting GitLab MCP Server with stdio transport
ERROR [2025-07-03 17:20:51,970] [pool-18-thread-1] [StdioClientTransport]: Error processing inbound message for line: Initializing server with transport mode: stdio
com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'Initializing': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (String)"Initializing server with transport mode: stdio"; line: 1, column: 13]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:2337)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:720)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._reportInvalidToken(ReaderBasedJsonParser.java:2903)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._reportInvalidToken(ReaderBasedJsonParser.java:2881)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._matchToken(ReaderBasedJsonParser.java:2662)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._handleOddValue(ReaderBasedJsonParser.java:1933)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:781)
	at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:4684)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4586)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3548)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3531)
	at io.modelcontextprotocol.spec.McpSchema.deserializeJsonRpcMessage(McpSchema.java:155)
[...]

PR #135, opened 2025-07-06
FIX: console to stderr

iwakitakuma33:
issue: #132

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -185,6 +185,19 @@ import {
 } from "./schemas.js";
 import { randomUUID } from "crypto";
 
+const pino = require('pino')
+const logger = pino({
+  level: process.env.LOG_LEVEL || 'info',
+  prettyPrint: {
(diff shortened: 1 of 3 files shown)


index.ts
```diff
@@ -185,6 +185,19 @@ import {
 } from "./schemas.js";
 import { randomUUID } from "crypto";
 
+const pino = require('pino')
```
fahrradflucht: This fails on my machine with:
```
const pino = require('pino');
             ^

ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and '<redacted>/gitlab-mcp/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

iwakitakuma33: @fahrradflucht 
thanks!!
I did fix it.

PR #151, opened 2025-07-09
tag with image push

iwakitakuma33:


iwakitakuma33:
@zereight 
Is it possible to use npm_push.sh for pushing from next time?
It is possible to release using the newest tag.

@bbang-dduck 
Is it possible to release v1.0.71 as a single release?
The version management of docker and npm package is not unified. I think it would be better to unify it with the release notes somewhere.

bbang-dduck (approved):

Issue #153, opened 2025-07-09
[Cursor 1.1.7] Invalid type for parameter 'merge_request_iid' in tool create_merge_request_thread


crespoxiao:
Title:
merge_request_iid type error: string/number both rejected in create_merge_request_thread

Body:
Hi,
When using MCP GitLab plugin with Cursor, all attempts to call create_merge_request_thread with merge_request_iid (as number or string, e.g. 763 or "763") result in the following error:
`Invalid type for parameter 'merge_request_iid' in tool create_merge_request_thread`

Possible root cause:
It seems to be a schema validation issue between the plugin and the ToolFormer API definition. Please help to check type handling for merge_request_iid, or if other param formatting has changed in new Cursor releases.

Thanks!

crespoxiao:
@iwakitakuma33 
hi ,i think it's the latest version 1.0.71

`xiaochengfei@Xiao-Chengfeis-MacBook-Pro OKKLine % npm list -g --depth=0 | grep mcp-gitlab

├── @zereight/mcp-gitlab@1.0.71
xiaochengfei@Xiao-Chengfeis-MacBook-Pro OKKLine % 
xiaochengfei@Xiao-Chengfeis-MacBook-Pro OKKLine % npm list -g | grep mcp-gitlab
├── @zereight/mcp-gitlab@1.0.71`

iwakitakuma33:
@crespoxiao 
It has been updated to a new version.
Please check again.

Please output the input that failed to the LLM.

crespoxiao:
@iwakitakuma33 
1.0.72 still has the issue with gpt-4.1, 

<img width="625" height="668" alt="Image" src="https://github.com/user-attachments/assets/fab792ba-b70e-4299-936d-4667d2f89715" />
<img width="618" height="617" alt="Image" src="https://github.com/user-attachments/assets/b4cf20f5-f547-47f7-826d-bfd55839ed49" />

iwakitakuma33:
@crespoxiao 
Please provide the input information that the LLM passed to the gitlab-mcp server when it failed.
Please provide text instead of images.

Please tell us which tool you used and how it failed.

crespoxiao:
@iwakitakuma33 

i'm using cursor1.1.7, llm is chatgpt-4.1 , chat box input is:
`review the code and add review comments directly to gitlab inline for  !763 of project id 6285`
mcp tool config and project info is correct because when use claude4, can successfully create summary comments. 


[...]

PR #161, opened 2025-07-11
FEAT: id is string or number

iwakitakuma33:
issue: #153


iwakitakuma33:
The documentation stated “integer,” but since there is an API that can handle uuid, it has been corrected to z.coerce.string.

zereight (approved):

Issue #186, opened 2025-07-28
No official way to "merge" a merge request via MCP tools?


mikegreiling:
I cannot seem to figure out a way to actually accept/merge a merge request using the MCP tools available in this library.

The standard GitLab API has a `PUT /projects/:id/merge_requests/:merge_request_iid/merge` endpoint, but this MCP server doesn't appear to implement it.

iwakitakuma33:
@mikegreiling 

I add PR.

could you try https://github.com/zereight/gitlab-mcp/pull/193 in your local with docker?

PR #193, opened 2025-07-31
FEAT: merge MR

iwakitakuma33:
issue: #186


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true
+}
```
zereight: Could you clarify why .vscode/settings.json was included in this PR?

zereight (changes requested):

.vscode/settings.json
iwakitakuma33: @zereight 
thanks.
that is not needed.

I deleted it.

Issue #199, opened 2025-08-06
merge_when_pipeline_succeeds incorrectly typed


Bowbaq:
I think #150 possibly broke the ability to set a MR to merge when the pipeline succeeds using the MCP with Cursor. I've tried to provide very explicit instructions to pass `merge_when_pipeline_succeeds` as a boolean, however Cursor(in Auto mode) insists on passing it as a string.

Request:
```
{
  "project_id": "Bowbaq/orderbot",
  "merge_request_iid": "2240",
  "merge_when_pipeline_succeeds": "True",
  "remove_source_branch": true,
  "squash": true
}
```

Result:
```
{
  "id": "404324268",
  "iid": "2240",
  "project_id": "8782292",
  "title": "Bump github.com/aws/aws-sdk-go-v2/service/s3 from 1.84.1 to 1.85.1",
  ...
  "merge_when_pipeline_succeeds": false, // <-- This should be true
  "squash": true,
  "labels": [
    "dependencies",
    "go"
  ]
}
```

This is what I have in my rules to try & coerce it to pass the correct type, which isn't working:
```
10. **Configure merge request settings via GitLab MCP**

- **Important**: All parameters should be proper boolean values (`true`/`false`), not strings. Use lowercase `true`/`false`, not `True`/`False`. Never use `"merge_when_pipeline_succeeds": "True"` or `"merge_when_pipeline_succeeds": "true"`
- **Important**: If the MCP call returns `"merge_when_pipeline_succeeds": false`, then the operation failed. Try again making sure to pass `"merge_when_pipeline_succeeds": true` as a **boolean** value.
- Use the GitLab MCP to update the merge request with the following settings:

{
   "merge_when_pipeline_succeeds": true,
   "remove_source_branch": true,
   "squash": true
}
```

If I manually ask the agent enough times to send a boolean, it sometimes does it & that works fine.

Bowbaq:
As far as I can tell, the API call is only successful if passing the boolean `true`. Passing `"true"` or `"True"` does not return an error, but also does not set the flag properly (ie. `merge_when_pipeline_succeeds` is `false` in the response).

The LLM seems quite capable of passing booleans for other fields in the request (in the example above: `remove_source_branch` and `squash` are both correctly passed as booleans).

Bowbaq:
The output above is from the `update_merge_request` tool

PR #201, opened 2025-08-08
FIX: flexible boolean

iwakitakuma33:
issue: #199

### DIFF
--- a/customSchemas.ts
+++ b/customSchemas.ts
@@ -1,12 +1,11 @@
 import { z } from "zod";
-import { pino } from 'pino';
+import { pino } from "pino";
 const DEFAULT_NULL = process.env.DEFAULT_NULL === "true";
 
-
 const logger = pino({
-  level: process.env.LOG_LEVEL || 'info',
+  level: process.env.LOG_LEVEL || "info",
   transport: {
-    target: 'pino-pretty',
+    target: "pino-pretty",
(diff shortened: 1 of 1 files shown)


customSchemas.ts
```diff
@@ -15,11 +14,19 @@ const logger = pino({
...
+  if (typeof val === "boolean") {
     return val;
+  }
+  let result = "false";
+  try {
+    result = String(val).toLowerCase();
+  } catch {
+    return false;
+  }
+  return result in ["true", "t", "1"];
```
zereight: Did you intend "include"?

Screenshot 2025-08-09 at 12 30 11 AM

customSchemas.ts
```diff
@@ -1,12 +1,11 @@
 import { z } from "zod";
-import { pino } from 'pino';
+import { pino } from "pino";
```
zereight: On second thought, it seems like it's better to put your prettier settings on a PR and share them together. I will follow you.

customSchemas.ts
iwakitakuma33: @zereight 
thanks.
I fixed it.

customSchemas.ts
iwakitakuma33: It seems that the wrong settings format was applied. I have reverted it back.

PR #208, opened 2025-08-13
feat: Add NPM publish workflow

zereight:
I created an npm workflow. Like docker yml.

And I added secret to the repo. like this
Screenshot 2025-08-13 at 6 52 57 PM

The DOCKERHUB_USERNAME is `zeregith050`.

I'd like to check it success.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,23 @@
+name: NPM Publish
+
+on:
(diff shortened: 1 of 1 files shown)


iwakitakuma33:
@zereight 
When pushing to my docker repo, it may be necessary to separate the login user and the push destination.

This is not necessary when pushing to your docker hub.

PR #222, opened 2025-08-19
Add `EXPOSE`  on Dockerfile

ardyfeb:
Add `EXPOSE` with default port on DockerFile


zereight (changes requested):
Could you please provide more information about why this change is needed and in what situations it would be helpful?

Issue #120, opened 2025-06-28
limit scope for all operations within given group


Guria:
I need to ensure all searches, requests, queries are forced to be limited to happen within my organization scope

iwakitakuma33:
@Guria what do you mean?
do you wanna use the one project name?

Guria:
We have all company code scoped at https://gitlab.com/company_name
I want to be able to define company_name or its group id in env variables to ensure all search will be limited to be in this group only.

Guria:
Yes.

iwakitakuma33:
@Guria 
I added it.

could you check in your local with docker.?
https://github.com/zereight/gitlab-mcp/pull/169

Guria:
@iwakitakuma33 
hmm, I'll try. though can you elaborate your solution? I am not sure it is actually what I am looking for. It looks like expects project id or path. Not group name. But probably I am just misled by function name

iwakitakuma33:
@Guria 
When GITLAB_PROJECT_ID is set, LLM will always use GITLAB_PROJECT_ID regardless of the value used.

The variable name may be incorrect.
When GITLAB_GROUP_NAME is set, restrict the API to specify GITLAB_GROUP_NAME/PROJECT_PATH when using the API.

Guria:
Here is what I actually looking for:
https://github.com/Adit-999/gitlab-mcp/blob/08d543e992bcc8a1e9f78963077beb696d0f83e8/src/services/search.py#L71

It is a slightly different endpoint

PR #228, opened 2025-08-24
feat: add group-scoped search tools

Guria:
related to #120 
it doesn't limit all tools usages but introduces relevant group-scoped tools which use corresponded api endpoints


Guria:
@iwakitakuma33 @zereight

PR #233, opened 2025-08-29
Add ability to trigger / retry / cancel individual jobs

jmif:


index.ts
```diff
@@ -845,6 +863,9 @@ const readOnlyTools = [
   "list_pipeline_trigger_jobs",
   "get_pipeline_job",
   "get_pipeline_job_output",
+  "play_pipeline_job",
+  "retry_pipeline_job",
+  "cancel_pipeline_job",
```
zereight: Why included Read Only Mode?

zereight (changes requested):

zereight:
Thx for contribute!

index.ts
jmif: Good catch, fixed

zereight (dismissed):

zereight:
@jmif Sorry, can you fix conflict?

PR #261, opened 2025-10-15
Support Oauth client authentication with Gitlab

kc9ddi:
This PR allows the MCP to use OAuth client authentication to authenticate with Gitlab. For many users (particularly team members using a company/team-wide Gitlab account), this may be an easier way to authenticate, instead of needing to create a personal access token and copy/paste it into their environment.

The approach used is,
1. The MCP stands up a lightweight HTTP server on 127.0.0.1, which serves the OAuth callback URL. If multiple MCP instances are started concurrently (such as can happen if running multiple Claude Code sessions in parallel), only a single HTTP server will be started, and will be reused across multiple MCP instances.
2. Based on configuration, the MCP will operate either with OAuth authentication, or with the existing access token authentication
3. If configured to use OAuth authentication, and a new token is needed, then the user's browser will open to the Gitlab OAuth authentication page when the MCP is launched
4. Once the authentication flow completes successfully, the user can use the MCP


zereight:
fix conflict plz!

zereight:
Thank you, this PR will be merged once the v2.0.9 release is stable.

zereight (approved):

PR #267, opened 2025-10-28
Fix regex for GitLab PAT token validation

gravis:
Some tokens have a "." (period) which makes them invalid from this mcp server perspective, returning a 401 while the token is actually valid


davidruzicka:
Thank you, GitLab tokens (job/deploy) use segmentation with dot.

PR #293, opened 2025-12-10
feat: allow global merge request listing (optional project_id)

oleksiigolikov:
Allows `list_merge_requests` to be called without a `project_id`.

When `project_id` is omitted, it returns all merge requests the authenticated user has access to (global scope).
When `project_id` is provided, it behaves as before (project scope).

Also added comprehensive tests using `MockGitLabServer` to verify both scenarios and filtering.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "test:server": "npm run build && node build/test/test-all-transport-server.js",
     "test:mcp:readonly": "tsx test/readonly-mcp-tests.ts",
     "test:oauth": "tsx test/oauth-tests.ts",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -974,7 +974,6 @@
       "resolved": "https://registry.npmjs.org/@types/node/-/node-22.13.14.tgz",
       "integrity": "sha512-Zs/Ollc1SJ8nKUAgc7ivOEdIBM8JAKgrqqUYi2J997JuKO7/tpQC+WCetQ1sypiKCQWHdvdg9wBNpUPEWZae7w==",
       "license": "MIT",
-      "peer": true,
```
zereight: Can you pull and `nvm use`?

PR #311, opened 2026-01-09
fix: Fix get_file_contents failing in Trae

yodagg:
- Environment : Self-hosted company GitLab 10.7.3; Trae IDE as MCP client
- Issue : `list_projects/get_project/get_repository_tree` worked, but `get_file_contents` failed with Invalid arguments: `Invalid input` , blocking reads of files like package.json
- Root cause :
  - Trae may send `path` instead of `file_path` , and/or omit `project_id` when `GITLAB_PROJECT_ID` is configured, causing strict input schema validation to reject the call
  - GitLab 10.7.3 file API responses can differ from the strict expected schema, causing response parsing failures
- Change (schemas only, `schemas.ts` ):
  - Make `GetFileContentsSchema` tolerant: accept path alias, allow optional project_id , normalize inputs
  - Relax `GitLabFileContentSchema` fields to improve compatibility with older/self-hosted GitLab responses


yodagg:
> Can you check some ai reviews?

OK

PR #327, opened 2026-01-20
fix: use getConfig for GITLAB_API_URL to support --api-url CLI argument

alfonsodg:
## Problem
The `GITLAB_API_URL` constant was reading directly from `process.env.GITLAB_API_URL`, ignoring the `--api-url` CLI argument.

## Solution
Changed line 1393 to use `getConfig('api-url', 'GITLAB_API_URL')` instead of `process.env.GITLAB_API_URL`, consistent with how other configuration options are handled.

## Testing
```bash
node build/index.js --token=xxx --api-url=https://gitlab.example.com/api/v4
```

Now correctly uses the CLI argument instead of requiring the environment variable.


zereight (changes requested):
change package name plz

zereight:
duplicated
https://github.com/zereight/gitlab-mcp/pull/349

PR #357, opened 2026-02-26
Housekeeping: close MCP resources on shutdown

seahyc:
## Summary\n- add client pool shutdown () to destroy keep-alive HTTP/HTTPS agents\n- invoke cleanup on stdio transport close\n- add SSE graceful shutdown path to close transports and pool\n- ensure streamable HTTP graceful shutdown also destroys pooled agents\n\n## Why\nPrevents lingering sockets/file descriptors after client disconnect or process shutdown.


zereight (changes requested):
nit: refind PR

PR #405, opened 2026-04-03
fix: auto-refresh OAuth token on 401 responses

YassineValue:
When GitLab invalidates an OAuth token server-side before the local expiry timer, API calls fail with 401 and users must manually reconnect. This adds a fetch-level interceptor that automatically force-refreshes the token and retries the request once on 401.

Changes:
- oauth.ts: Add `force` parameter to getAccessToken() to bypass local expiry check
- index.ts: Add wrapWithAuthRetry() interceptor that wraps fetch at every assignment point (initial + cookie reloads)

The retry calls baseFetch directly (not the wrapper), making infinite loops impossible. Non-OAuth mode is unaffected.


zereight (changes requested):
plz check comments!

medyas:
done

zereight (approved):

PR #424, opened 2026-04-17
fix(oauth): correctly reference endpoints during discovery

Killusions:
Currently this only works if the server is served at a domain root


Killusions:
@zereight Currently the remote oauth server/proxy mode only works if deployed at the root path, we'd like to deploy it at a path

zereight (approved):

PR #434, opened 2026-04-20
feat(create_repository): support namespace_id for group/user targeting

awkto:
## Summary

`create_repository` currently always creates projects in the authenticated user's personal namespace — there is no way to create a project under a group, or under another user (for admins). This PR adds an optional `namespace_id` field that passes straight through to `POST /projects`, so agents can target groups like `gitlab.example.com/my-team/<repo>` in one call.

When `namespace_id` is omitted, behavior is unchanged (defaults to the authed user's personal namespace).

Accepts either a numeric namespace ID or a full path string, matching the GitLab REST API's own flexibility for this field.

## Changes

- `schemas.ts`
  - Add `namespace_id` (optional, `number | string`) to `CreateRepositorySchema` (the MCP tool input schema exposed via `tools/registry.ts`).
  - Add the same field to `CreateRepositoryOptionsSchema` (the internal function param type).
- `index.ts` — include `namespace_id` in the `POST /projects` body only when it is provided (conditional spread, so the payload shape is untouched for the default case).

## Test plan

- [x] `npm install` (which runs `prepare` → `tsc`) succeeds with no type errors.
- [ ] Call `create_repository` without `namespace_id` → project is created in the authed user's personal namespace (existing behavior).
- [ ] Call `create_repository` with `namespace_id: <group_id>` → project is created under that group.
- [ ] Call `create_repository` with `namespace_id: "my-group/sub-group"` → project is created under that path.
- [ ] Call with a namespace the user doesn't have write permission to → GitLab returns 403; the MCP surfaces the error (no change to error handling).

## Motivation

Came up while using the MCP against a self-hosted GitLab: the agent needed to create projects under a dedicated group (e.g. `github/<repo>` for GitHub-mirrored projects), and the only workaround was to shell out to `glab api` with `-f namespace_id=<id>`. One-line addition on the server side removes the need for that workaround entirely.

### DIFF
--- a/schemas.ts
+++ b/schemas.ts
@@ -746,6 +746,12 @@ export const CreateRepositoryOptionsSchema = z.object({
   description: z.string().optional(),
   visibility: z.enum(["private", "internal", "public"]).optional(), // Changed from private to match GitLab API
   initialize_with_readme: z.coerce.boolean().optional(), // Changed from auto_init to match GitLab API
+  namespace_id: z
+    .union([z.coerce.number(), z.string()])
+    .optional()
+    .describe(
+      "Namespace ID (numeric) or full path for the group/user to create the project under. Omit to use the authenticated user's personal namespace."
+    ),
 });
 
 export const CreateIssueOptionsSchema = z.object({
@@ -1294,6 +1300,12 @@ export const CreateRepositorySchema = z.object({
     .optional()
     .describe("Repository visibility level"),
   initialize_with_readme: z.coerce.boolean().optional().describe("Initialize with README.md"),
+  namespace_id: z
+    .union([z.coerce.number(), z.string()])
+    .optional()
+    .describe(
+      "Namespace ID (numeric) or full path to create the project under. Omit to use the authenticated user's personal namespace."
+    ),
 });
 
 export const GetFileContentsSchema = z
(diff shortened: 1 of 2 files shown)


schemas.ts
```diff
@@ -1294,6 +1300,12 @@ export const CreateRepositorySchema = z.object({
     .optional()
     .describe("Repository visibility level"),
   initialize_with_readme: z.coerce.boolean().optional().describe("Initialize with README.md"),
+  namespace_id: z
+    .union([z.coerce.number(), z.string()])
```
zereight, line 1304: I think we should narrow this to numeric namespace IDs unless we also add a lookup step. GitLab's `POST /projects` API documents `namespace_id` as an integer group/subgroup ID, but this schema and the tool description accept a full path string and pass it straight through to the create-project request. That means inputs like `"my-group/sub-group"` are advertised as valid by the MCP schema, but they are not actually supported by this endpoint contract.

Could we either make this `z.coerce.number().optional()` and update the descriptions/test plan to say numeric namespace ID only, or resolve a full namespace path to its numeric ID before calling `POST /projects`?

zereight (commented):
Left one comment about the `namespace_id` input contract.

zereight (changes requested):
Requesting changes because the schema currently advertises full namespace path strings for `namespace_id`, but the GitLab `POST /projects` contract expects a numeric namespace ID. Please either narrow the schema/descriptions/test plan to numeric IDs, or add a namespace path lookup before creating the project.

zereight:
Closing this because the requested changes have been stale for an extended period and there is no actionable re-review signal at the moment.\n\nIf you want to continue this work, please reopen the PR or open a fresh one with the requested fixes addressed, and we can review it again.

PR #435, opened 2026-04-23
expose GitLab tags API

tequeter:
- Port PR #211 / commit afaebf8 to the current main branch.
- Improve tool descriptions.
- Integrate the feature as a toolset.
- Reference the feature in the README.


schemas.ts
```diff
@@ -3031,6 +3031,80 @@ export const DownloadReleaseAssetSchema = z.object({
...
+  release: z
+    .object({
+      tag_name: z.string(),
+      description: z.string(),
+    })
+    .nullable(),
+  protected: z.boolean(),
+});
+
+export const GitLabTagSignatureSchema = z.object({
```
zereight, line 3112: This schema does not match the response returned by GitLab's tag signature endpoint. `GET /projects/:id/repository/tags/:tag_name/signature` returns `signature_type`, `verification_status`, and an `x509_certificate` object, but this schema requires `signature` and `public_key`. A valid signed-tag response from GitLab will fail zod parsing here, so `get_tag_signature` errors instead of returning the signature metadata.

Suggested shape:

```ts
export const GitLabTagSignatureSchema = z.object({
  signature_type: z.literal("X509"),
  verification_status: z.string(),
  x509_certificate: z.object({
    id: z.number(),
    subject: z.string(),
    subject_key_identifier: z.string(),
    email: z.string().nullable().optional(),
    serial_number: z.number(),
    certificate_status: z.string(),
    x509_issuer: z.object({
      id: z.number(),
      subject: z.string(),
      subject_key_identifier: z.string(),
      crl_url: z.string().nullable().optional(),
    }),
  }),
});
```

If we want to be more forward-compatible, the nested certificate fields could be `.passthrough()`, but the important part is to model `verification_status` and `x509_certificate` instead of `signature` / `public_key`.

schemas.ts
```diff
@@ -3031,6 +3031,80 @@ export const DownloadReleaseAssetSchema = z.object({
...
+  project_id: z.coerce.string().describe("Project ID or URL-encoded path"),
+  tag_name: z.string().describe("The name of the tag"),
+});
+
+export const CreateTagSchema = z.object({
+  project_id: z.coerce.string().describe("Project ID or URL-encoded path"),
+  tag_name: z.string().describe("The name of the tag"),
+  ref: z.string().describe("Create tag using commit SHA, another tag name, or branch name"),
+  message: z.string().optional().describe("Create annotated tag with message"),
+  release_description: z.string().optional().describe("Release notes for the tag"),
```
zereight: `release_description` is not part of GitLab's `POST /projects/:id/repository/tags` contract. The tag endpoint accepts `tag_name`, `ref`, and optional `message`; passing `release_description` here advertises release-note behavior that this endpoint does not implement. Depending on GitLab's behavior, the field is either ignored or rejected, but in either case the MCP tool contract is misleading.

Suggested shape:

```ts
export const CreateTagSchema = z.object({
  project_id: z.coerce.string().describe("Project ID or URL-encoded path"),
  tag_name: z.string().describe("The name of the tag"),
  ref: z.string().describe("Create tag using commit SHA, another tag name, or branch name"),
  message: z.string().optional().describe("Create annotated tag with message"),
});
```

If creating release notes is intended, that should be a separate call to the Releases API after tag creation, not an extra field sent to the tags endpoint.

zereight (changes requested):
Requesting changes because two exposed tag tool contracts do not match the GitLab Tags API: `get_tag_signature` rejects valid signature responses, and `create_tag` advertises a release-description field that the tags endpoint does not support.

tequeter:
Thank you for the review, @zereight!

I added 9a8469be which directly addresses your concerns, then rebased the two commits onto the latest main (not fast-foward because we both edited the test list in package.json).

zereight:
Thanks, this addresses my review comments.

zereight (approved):
Approved after re-review. The requested tag API contract fixes are in place, and the PR checks are green.

PR #436, opened 2026-04-23
feat: add OAuth callback proxy mode (GITLAB_OAUTH_CALLBACK_PROXY)

PedroFCM:
## Problem

When `GITLAB_MCP_OAUTH=true`, the MCP server passes the client's callback URL straight through to GitLab. Every MCP client's callback URL must be pre-registered in the GitLab OAuth Application settings, requiring GitLab admin involvement for each new client.

## Solution

New opt-in mode: `GITLAB_OAUTH_CALLBACK_PROXY=true`. The MCP server intercepts the OAuth callback from GitLab, exchanges the code for tokens server-side, and redirects to the client with a proxy code. Only one fixed callback URL (`${MCP_SERVER_URL}/callback`) needs to be registered in GitLab.

Disabled by default — existing passthrough behavior is unchanged.

## How it works

1. Client calls `/authorize` → server stores client params, generates its own PKCE pair, redirects to GitLab with the fixed callback URL
2. User authorizes on GitLab → GitLab redirects to the MCP server's `/callback`
3. Server exchanges the code with GitLab, stores tokens, generates a proxy code, redirects to the client
4. Client calls `/token` with the proxy code → server verifies client PKCE, returns stored tokens

See `docs/oauth-callback-proxy.md` for sequence diagrams and full documentation.

## Security

- Dual PKCE: separate pairs for client↔server and server↔GitLab
- Client PKCE verification: `code_verifier` mandatory when `code_challenge` was stored
- One-time-use proxy codes with 10-min TTL
- State replay protection
- Sanitized error responses
- Bounded in-memory LRU storage (max 1000 entries)

## Changes

- `oauth-proxy.ts`: callback proxy logic, `BoundedLRUMap<V>` replaces `BoundedClientCache`
- `config.ts`: new `GITLAB_OAUTH_CALLBACK_PROXY` env var
- `index.ts`: mounts path-aware `/callback` route
- `docs/oauth-callback-proxy.md`: documentation with Mermaid sequence diagrams
- `test/callback-proxy-tests.ts`: 7 tests (happy path, one-time-use, PKCE, state replay, dual PKCE)

## Testing

- All 7 new tests pass
- Tested end-to-end with `mcp-remote` and a Streamable HTTP client against a live GitLab instance
- Existing tests unaffected

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -82,6 +82,8 @@ export const GITLAB_OAUTH_SCOPES =
   GITLAB_OAUTH_SCOPES_RAW
     ? GITLAB_OAUTH_SCOPES_RAW.split(",").map((s) => s.trim()).filter(Boolean)
     : undefined;
+export const GITLAB_OAUTH_CALLBACK_PROXY =
+  getConfig("oauth-callback-proxy", "GITLAB_OAUTH_CALLBACK_PROXY") === "true";
 export const ENABLE_DYNAMIC_API_URL =
   getConfig("enable-dynamic-api-url", "ENABLE_DYNAMIC_API_URL") === "true";
 
--- a/docs/oauth-callback-proxy.md
+++ b/docs/oauth-callback-proxy.md
@@ -1,0 +1,107 @@
+# GitLab MCP OAuth Callback Proxy
+
+## Problem
+
+The current OAuth flow requires each MCP client's callback URL to be pre-registered in the GitLab OAuth Application settings. This means GitLab admin involvement for every new client deployment.
+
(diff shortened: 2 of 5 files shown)


danielmcorreia:
Having to pre-register every client callback URL is a current pain point and doesn’t scale well when you’re setting up new clients.

Really like this proxy approach, feels like a clean way to remove that bottleneck. Thanks for putting this together @PedroFCM !

zereight (approved):
Approved. client_id/redirect_uri binding hardening can be handled in a follow-up without changing the env or user-facing setup.

PR #438, opened 2026-04-26
fix: defensive guards for project_id coercion, labels query, and null MR response

alfonsodg:
## Summary

Three defensive fixes for edge cases encountered in production:

### 1. Guard `getEffectiveProjectId` against coerced string literals

`z.coerce.string()` converts `undefined`/`null` to the literal strings `"undefined"`/`"null"`, which bypass the existing `!projectId` falsy check. This causes the function to treat them as real project IDs, leading to 404 errors from the GitLab API.

**Fix:** Early return when projectId is falsy or equals `"undefined"`/`"null"`, falling back to allowed project IDs or `GITLAB_PROJECT_ID`.

### 2. Fix labels query parameter format in `listIssues`

The GitLab API expects labels as a comma-separated string (`labels=bug,feature`), not array notation (`labels[]=bug&labels[]=feature`). The array format causes the API to ignore the filter silently.

Ref: https://docs.gitlab.com/api/issues.html#list-issues

**Fix:** Use `value.join(",")` for labels while keeping `[]` notation for `assignee_username` which does expect array format.

### 3. Null coalescing for `listMergeRequests` response

When `listMergeRequests` returns `null` (e.g., empty project or API edge case), `JSON.stringify(null)` produces the string `"null"` instead of a valid array.

**Fix:** `mergeRequests ?? []` ensures a valid JSON array is always returned.

## Testing

- TypeScript compilation passes (`tsc --noEmit`)
- Tested against self-hosted GitLab 17.x with multiple project configurations

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1281,6 +1281,22 @@ async function handleGitLabError(response: import("node-fetch").Response): Promi
  * @throws {Error} If GITLAB_ALLOWED_PROJECT_IDS is set and the requested project is not in the whitelist
  */
 function getEffectiveProjectId(projectId: string): string {
+  // Guard against z.coerce.string() converting undefined/null to literal strings
+  if (!projectId || projectId === "undefined" || projectId === "null") {
+    if (GITLAB_ALLOWED_PROJECT_IDS.length === 1) {
+      return GITLAB_ALLOWED_PROJECT_IDS[0];
+    }
+    if (GITLAB_ALLOWED_PROJECT_IDS.length > 1) {
+      throw new Error(
+        `Multiple projects allowed (${GITLAB_ALLOWED_PROJECT_IDS.join(", ")}). Please specify a project ID.`
+      );
+    }
+    if (GITLAB_PROJECT_ID) {
+      return GITLAB_PROJECT_ID;
+    }
+    throw new Error("No project ID provided and GITLAB_PROJECT_ID is not set");
+  }
+
   if (GITLAB_ALLOWED_PROJECT_IDS.length > 0) {
     // If there's only one allowed project, use it as default
     if (GITLAB_ALLOWED_PROJECT_IDS.length === 1 && !projectId) {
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -9390,7 +9400,7 @@ async function handleToolCall(params: any) {
 
         const mergeRequests = await listMergeRequests(project_id, cleanedOptions);
         return {
-          content: [{ type: "text", text: JSON.stringify(mergeRequests, null, 2) }],
+          content: [{ type: "text", text: JSON.stringify(mergeRequests ?? [], null, 2) }],
```
zereight, line 9403: This does not actually handle the `null` response case described in the PR. `listMergeRequests()` still parses the response with `z.array(GitLabMergeRequestSchema).parse(data)` before returning, so if GitLab returns `null`, it will throw inside `listMergeRequests()` and this `mergeRequests ?? []` fallback is never reached.

Could we move the fallback into `listMergeRequests()` itself, for example by normalizing `data ?? []` before the zod array parse, or by changing the parser to explicitly accept `null` and return `[]` there? That keeps the behavior consistent for every caller instead of only this tool-call branch.

zereight (commented):
Left one comment about the null merge-request response handling.

index.ts
zereight: Suggested shape:

```ts
const data = await response.json();
const normalizedData = data ?? [];
return z.array(GitLabMergeRequestSchema).parse(normalizedData);
```

If we only want to tolerate `null` and still reject other unexpected shapes, this keeps the existing zod validation intact while making the documented `null -> []` behavior actually reachable for every `listMergeRequests()` caller.

zereight (changes requested):
.

index.ts
zereight: Still unresolved — this fallback is behind the Zod parse step, so a null response still throws before reaching `mergeRequests ?? []`. The suggested `data ?? []` normalization inside `listMergeRequests()` would make the fix reachable for all callers.

zereight (commented):
**Blocking (1) — existing thread at line 9403**

The `mergeRequests ?? []` fallback is placed behind Zod parse step in `listMergeRequests()`, so a null API response still throws before reaching it. Same assessment as zereight's earlier feedback — move `data ?? []` normalization into `listMergeRequests()` before the `.parse()` call.

**Non-blocking notes**

1. **Empty labels array**: `labels.join(",")` on `[]` appends `labels=` which is new behavior. Consider skipping when empty.
2. **Tests**: The PR would benefit from regression tests for the three edge cases — coerced project IDs, labels query serialization, and null MR response.
3. **getEffectiveProjectId placement**: The early guard duplicates some logic below. Schema-level normalization (preprocess null→undefined) would be cleaner, but the current fix is pragmatic.

**Clean**
- `labels` comma-separated is the correct format per GitLab API docs.
- `assignee_username[]` notation correctly preserved.
- No auth/transport/security boundary impact.

zereight:
Closing this because the requested changes have been stale for an extended period and there is no actionable re-review signal at the moment.\n\nIf you want to continue this work, please reopen the PR or open a fresh one with the requested fixes addressed, and we can review it again.

PR #442, opened 2026-04-30
feat: stateless OAuth / session mode for multi-pod HPA deployments

### LINKED ISSUE #441: OAuth and session state not shared across pods breaks HPA deployments
## Summary

When `@zereight/mcp-gitlab` runs with `STREAMABLE_HTTP=true` behind a
Kubernetes deployment with multiple replicas (HPA), OAuth flows and MCP
sessions fail whenever consecutive requests of the same logical session are
routed to different pods. All per-session state is held in per-process
memory with no shared-store abstraction and no way to survive a pod hop.

This affects `REMOTE_AUTHORIZATION=true` and every `GITLAB_MCP_OAUTH=true`
flow (both passthrough and callback-proxy modes).

## Affected version

- `@zereight/mcp-gitlab` at current `main` (2.1.4)
- Node 18+, Streamable HTTP transport

## Preconditions

- `STREAMABLE_HTTP=true`
- Multiple replicas of the server (any `HorizontalPodAutoscaler`, or any
  round-robin / least-conn load balancer in front of ≥2 processes)
- One of:
  - `REMOTE_AUTHORIZATION=true`
  - `GITLAB_MCP_OAUTH=true` (passthrough mode)
  - `GITLAB_MCP_OAUTH=true` + `GITLAB_OAUTH_CALLBACK_PROXY=true`

Traefik's built-in stickiness is cookie-based, not header-based, so
`Mcp-Session-Id` header stickiness is not natively available and does not
cover the OAuth callback endpoints anyway (they carry no MCP header).

## Reproduction

```bash
# Pod A and Pod B — same image, same config, different processes
OAUTH_VARS="STREAMABLE_HTTP=true GITLAB_MCP_OAUTH=true MCP_SERVER_URL=... GITLAB_OAUTH_APP_ID=... GITLAB_API_URL=..."

# Passthrough-mode failure
curl -X POST http://pod-A:3002/register    -d '{"redirect_uris":["https://client/cb"]}'   # returns client_id=X
curl -i  http://pod-B:3002/authorize?client_id=X&redirect_uri=...                          # 400 "Unregistered redirect_uri"

# REMOTE_AUTHORIZATION failure
# Init on A returns Mcp-Session-Id=S.
# Subsequent call on B with header Mcp-Session-Id: S → 401 "Missing Private-Token…"

# Callback-proxy failure
# /authorize on A stores pending state P in memory.
# GitLab redirects browser to /callback on B (round-robin) → 400 "Unknown or expired state parameter"
```

## Root cause

Five distinct in-memory stores on the serving pod:

| File | Symbol | Lifetime | Purpose |
|---|---|---|---|
| `oauth-proxy.ts` | `_clientCache: BoundedLRUMap` | DCR entry | written on `/register`, read on `/authorize` + `/token` |
| `oauth-proxy.ts` | `_pendingAuth: BoundedLRUMap` | 10 min | written on `/authorize`, read on `/callback` (callback-proxy mode) |
| `oauth-proxy.ts` | `_storedTokens: BoundedLRUMap` | 10 min | written on `/callback`, read on `/token` (callback-proxy mode) |
| `index.ts` | `authBySession: Record<sid, AuthData>` | `SESSION_TIMEOUT_SECONDS` | every `/mcp` request consults this |
| `index.ts` | `streamableTransports: Record<sid, Transport>` | connection | SDK transport bound to a live socket |

Each is a per-process `Map` / `Record`. Pods have no way to share them.

## Impact

- Every MCP client connecting to a multi-pod deployment fails on the first
  cross-pod hop for all three auth modes above.
- Rolling deploys, scale-in, and routine pod restarts silently invalidate
  in-flight sessions and OAuth consent windows.
- Observable symptoms: intermittent 400 / 401 responses that cannot be
  reproduced on a single-replica local deployment.

## Why header-based stickiness is not a sufficient fix

- Traefik's native stickiness is cookie-based, not header-based; rewriting
  `Mcp-Session-Id` into a sticky cookie is non-trivial and only some MCP
  clients honour cookies.
- The first `/mcp` request that creates the session carries no
  `Mcp-Session-Id` header yet — the LB cannot hash on a value the server
  is about to mint.
- Browser-driven OAuth `/callback` requests carry no `Mcp-Session-Id`
  header at all.
- Pod restart / HPA scale-in invalidates affinity regardless: the owning
  pod is gone, so stickiness points at nothing.

Shared cache (Redis / PVC) would solve it but introduces an external
dependency. A cheaper option exists.

## Proposed fix (see linked PR) — stateless opaque tokens

Encode every piece of per-session state into the opaque OAuth values
themselves (`client_id`, OAuth `state`, OAuth `code`, `Mcp-Session-Id`)
using HMAC or AEAD keyed off a single server-side secret.

- Public payloads (DCR) use HMAC (signed).
- Secret payloads (PKCE verifier, bearer tokens) use AES-256-GCM (sealed).
- Per-purpose subkeys via HKDF from one env-var secret.
- Any pod holding the secret can verify/reconstruct state with no shared
  store and no sticky sessions.
- Backwards compatible: off by default behind `OAUTH_STATELESS_MODE=true`.

fabiogermann:
Closes #441

## Disclosure
This PR includes code and documentation co-authored with AI assistance

## Summary

Adds an opt-in `OAUTH_STATELESS_MODE` that makes the OAuth proxy, DCR
registry, and `Mcp-Session-Id` path safe to distribute across multiple
pods with **no external cache, no sticky sessions, and no affinity
requirement**. All per-session state travels on the wire as a signed or
AEAD-sealed opaque value, keyed off a single shared secret.

Default behaviour (`OAUTH_STATELESS_MODE=false`) is bit-identical to
before. Single-replica and stdio deployments are unaffected.

## The problem (see linked issue)

Today's code holds five in-memory stores per pod:

| Surface | Today | Problem under HPA |
|---|---|---|
| DCR `_clientCache` | `BoundedLRUMap` on pod | `/register` on A, `/authorize` on B ⇒ 400 |
| Callback-proxy `_pendingAuth` | `BoundedLRUMap` on pod | `/authorize` on A, `/callback` on B ⇒ 400 |
| Callback-proxy `_storedTokens` | `BoundedLRUMap` on pod | `/callback` on A, `/token` on B ⇒ 400 |
| `/mcp` `authBySession` | `Record<sid, AuthData>` on pod | init on A, tool call on B ⇒ 401 |
| Rate-limit counters | per-pod | N× configured limit under LB |

Header-based stickiness on `Mcp-Session-Id` does not fix it: Traefik is
cookie-based, the header is absent on init and on all OAuth callback
endpoints, and pod restart / scale-in invalidates affinity regardless.

## Fix

Encode each piece of per-session state into the opaque OAuth value
itself, using HMAC or AEAD keyed off a per-purpose subkey derived from
one master secret:

```
signed:   v1.<purpose>.<b64url(payload)>.<b64url(hmac)>
sealed:   v1.<purpose>.<b64url(nonce || ciphertext || tag)>
```

| Surface | After |
|---|---|
| DCR `client_id` | Signed (`v1.cid.…`) — public payload |
| OAuth `state` | Sealed (`v1.ps.…`) — contains proxy PKCE verifier |
| Proxy `code` | Sealed (`v1.pc.…`) — contains GitLab tokens |
| `Mcp-Session-Id` | Sealed (`v1.sid.…`) — contains bearer token |
| Rate-limit | Disabled in stateless mode (documented); apply upstream |

Cryptography:
- HMAC-SHA256 (signed) or AES-256-GCM (sealed), Node built-in `crypto`
- Per-purpose subkeys via HKDF-SHA256 — a token minted for one purpose
  cannot be verified as another
- 12-byte nonce, 16-byte tag, purpose as AEAD AAD
- Two-slot keyring for rotation
  (`OAUTH_STATELESS_SECRET` + `OAUTH_STATELESS_SECRET_PREVIOUS`)

## What reviewers should look at

- `stateless/` — new module (codec, secret, four purpose-specific helpers)
- `oauth-proxy.ts` — clientsStore, `authorize()`, `handleCallback()`,
  `exchangeAuthorizationCode()` each gain a stateless branch; legacy
  BoundedLRUMap path retained
- `index.ts` — new `handleStatelessMcpRequest` helper short-circuits
  ahead of the legacy `authBySession` / `streamableTransports` logic
- `config.ts` — six new env vars
- `docs/stateless-mode.md` — operator guide with threat model and
  rotation runbook

## Configuration

```bash
OAUTH_STATELESS_MODE=true
OAUTH_STATELESS_SECRET=$(openssl rand -base64 32)    # same on every pod
# optional:
OAUTH_STATELESS_SECRET_PREVIOUS=...                  # rotation
OAUTH_STATELESS_CLIENT_TTL_SECONDS=86400             # default 24h
OAUTH_STATELESS_PENDING_TTL_SECONDS=600              # default 10min
OAUTH_STATELESS_STORED_TTL_SECONDS=600               # default 10min
OAUTH_STATELESS_SESSION_TTL_SECONDS=3600             # defaults to SESSION_TIMEOUT_SECONDS
```

In Kubernetes, mount `OAUTH_STATELESS_SECRET` from a `Secret` resource
identical across all pods.

## Security model

| Value | Compromise impact | Mitigation |
|---|---|---|
| `client_id` | None (public) | — |
| OAuth `state` | Useless alone (GitLab code is single-use) | 10-min TTL |
| Proxy `code` | Contains GitLab access token | 10-min TTL + PKCE `code_verifier` check at `/token` |
| `Mcp-Session-Id` | Equivalent to stolen bearer token | TLS, log redaction (added in this PR), inactivity TTL |
| `OAUTH_STATELESS_SECRET` | Total forgery | Treat as bearer secret, K8s Secret with access audit |

Rotation: set `OAUTH_STATELESS_SECRET_PREVIOUS=<old>`, deploy, remove
after `max(TTL)` elapses. Omitting `_PREVIOUS` invalidates all
outstanding tokens immediately — the intended emergency response.

## Backward compatibility

- `OAUTH_STATELESS_MODE=false` (default): zero behaviour change. The
  existing BoundedLRUMap and `authBySession` paths are retained and
  selected at runtime.
- Legacy UUID client_ids and session_ids still flow through the legacy
  stubs in the provider, so pre-existing GitLab app UIDs and any
  mid-flight sessions from older pods continue to work as before.
- No new runtime dependencies. All crypto uses Node's built-in module.


index.ts
```diff
@@ -10048,6 +10147,173 @@ async function startStreamableHTTPServer(): Promise<void> {
...
+    if (freshAuthPresent) {
+      freshSid = mintSessionId(material, {
+        header: effective.header,
+        token: effective.token,
+        apiUrl: effective.apiUrl,
+      });
+      metrics.statelessSidRotated++;
+    } else if (incomingSid && looksLikeStatelessSessionId(incomingSid)) {
+      freshSid = incomingSid;
+    } else {
```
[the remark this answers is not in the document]
fabiogermann: addressed in: https://github.com/zereight/gitlab-mcp/pull/442/commits/3e7dbbe0e003bc38e38f205a8b763767c11e1fc9

stateless/secret.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Load the keyring from the environment. Returns null when stateless mode is
+ * disabled. Throws StatelessConfigError when enabled but misconfigured.
+ */
+export function loadKeyMaterialFromEnv(
+  env: NodeJS.ProcessEnv = process.env
+): StatelessKeyMaterial | null {
+  const enabled = (env.OAUTH_STATELESS_MODE ?? "").toLowerCase() === "true";
+  if (!enabled) return null;
```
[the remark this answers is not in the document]
fabiogermann: addressed in: https://github.com/zereight/gitlab-mcp/pull/442/commits/e3ab1b74714d16d54a672252ab237d820e9930ab

index.ts
```diff
@@ -10048,6 +10151,180 @@ async function startStreamableHTTPServer(): Promise<void> {
...
+
+    if (!effective) {
+      metrics.authFailures++;
+      metrics.statelessAuthFailures++;
+      res.status(401).json({
+        error: "Authentication required",
+        message:
+          "Stateless mode: provide Private-Token, JOB-TOKEN, or Authorization " +
+          "header, or a valid Mcp-Session-Id from a previous response.",
+      });
```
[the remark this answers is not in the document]
fabiogermann: addressed in: https://github.com/zereight/gitlab-mcp/pull/442/commits/8979b124498d8a7613edd265b5d1ea038ca7460f

index.ts
```diff
@@ -10048,6 +10151,180 @@ async function startStreamableHTTPServer(): Promise<void> {
...
+          apiUrl: GITLAB_API_URL,
+        };
+        freshAuthPresent = true;
+      }
+    }
+    if (freshAuthPresent) metrics.statelessAuthFromHeader++;
+
+    // Fall back to the sealed sid when no live headers are present.
+    const incomingSid = req.headers["mcp-session-id"] as string | undefined;
+    if (!effective && incomingSid && looksLikeStatelessSessionId(incomingSid)) {
```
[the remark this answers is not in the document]
fabiogermann: addressed in: https://github.com/zereight/gitlab-mcp/pull/442/commits/edbf1dcd9706a1ab33a3881c0c3bb7000b7949ee

index.ts
```diff
@@ -10173,11 +10522,29 @@ async function startStreamableHTTPServer(): Promise<void> {
 
   // Streamable HTTP endpoint - handles both session creation and message handling
   app.post("/mcp", mcpBearerAuth, async (req: Request, res: Response) => {
```
zereight, line 10561: This route still runs through `mcpBearerAuth` before `handleStatelessMcpRequest` gets a chance to open the sealed `Mcp-Session-Id`. In `GITLAB_MCP_OAUTH=true`, a normal subsequent `/mcp` request may only carry `Mcp-Session-Id` after initialization. With no `Private-Token`, `JOB-TOKEN`, or `Authorization` header, `mcpBearerAuth` falls through to `oauthBearerAuth` and returns 401 before the stateless handler can reconstruct auth from the sid.

That means the new stateless session path works for `REMOTE_AUTHORIZATION`, but the OAuth deployment mode described by this PR still fails for sid-only follow-up requests across pods. Could we bypass the bearer middleware when stateless mode is enabled and a session id is present, or move the stateless branch ahead of this middleware so it can validate/open the sid itself?

zereight (changes requested):
Requesting changes because the stateless `/mcp` path is still behind the OAuth bearer middleware, so `GITLAB_MCP_OAUTH=true` clients cannot use a sealed `Mcp-Session-Id` as the sole credential on subsequent requests. That breaks the main multi-pod stateless-session contract for OAuth deployments.

index.ts
zereight: Suggested shape in pseudocode:

```ts
function hasStatelessSessionId(req: Request): boolean {
  return Boolean(
    OAUTH_STATELESS_MODE &&
      STATELESS_MATERIAL &&
      readMcpSessionIdHeader(req)
  );
}

const mcpBearerAuth = GITLAB_MCP_OAUTH
  ? (req, res, next) => {
      // Let the stateless /mcp handler validate/open the sid.
      // That handler can then return 200 + rotated sid, or 404 for
      // expired/invalid sid, instead of failing early with OAuth 401.
      if (hasStatelessSessionId(req)) {
        next();
        return;
      }

      if (hasPrivateTokenOrJobToken(req)) {
        const authData = parseAuthHeaders(req);
        if (authData) {
          next();
          return;
        }

        res.status(401).json({ error: "Invalid Private-Token or JOB-TOKEN header" });
        return;
      }

      oauthBearerAuth(req, res, next);
    }
  : (_req, _res, next) => next();
```

I would intentionally key this off `Mcp-Session-Id` presence, not `looksLikeStatelessSessionId()`, so malformed/expired/legacy session ids still reach the stateless handler and get the intended `404 Session not found` response.

A regression test should cover `GITLAB_MCP_OAUTH=true` + stateless enabled: initialize with `Authorization`, capture `Mcp-Session-Id`, then send a follow-up `/mcp` request with only that sid and assert `200` plus a rotated sid.

fabiogermann: Great catch @zereight, I will adapt the implementation to fix that case and add appropriate tests to cover it.

fabiogermann: Pushed an update that should resolve the issue you raised.

config.ts
```diff
@@ -87,6 +87,49 @@ export const GITLAB_OAUTH_CALLBACK_PROXY =
...
+  "OAUTH_STATELESS_STORED_TTL_SECONDS",
+  "oauth-stateless-stored-ttl",
+  600
+);
+
+/** Defaults to SESSION_TIMEOUT_SECONDS when unset. */
+export const OAUTH_STATELESS_SESSION_TTL_SECONDS = _intEnv(
+  "OAUTH_STATELESS_SESSION_TTL_SECONDS",
+  "oauth-stateless-session-ttl",
+  Number.parseInt(
```
zereight: This can still disable stateless session expiry when `OAUTH_STATELESS_SESSION_TTL_SECONDS` is unset and `SESSION_TIMEOUT_SECONDS` is invalid. In that state, this fallback becomes `NaN`; `_intEnv()` returns the fallback directly when the stateless-specific value is missing, so `OAUTH_STATELESS_SESSION_TTL_SECONDS` becomes `NaN` instead of falling back to a safe positive default. That breaks the documented inactivity timeout for sealed `Mcp-Session-Id` values.

Could we parse `SESSION_TIMEOUT_SECONDS` through the same finite-positive guard, for example by computing a sanitized `SESSION_TIMEOUT_SECONDS` first and passing that value here, or by making `_intEnv()` validate its fallback before returning it?

zereight (changes requested):
Follow-up review: the OAuth sid-only middleware issue appears fixed in the latest head, but there is still one blocking stateless TTL fallback issue.

config.ts
fabiogermann: `MEOUT_SECONDS` now flows through `_intEnv` so it's guaranteed finite-positive before being passed as a fallback.  `_intEnv` itself validates its fallback before returning (with a load-time throw if no valid fallback exists).

zereight (commented):
Thanks!

zereight (approved):
Thanks, this addresses my review comments.

PR #457, opened 2026-05-07
fix(oauth-callback-proxy): serve at static route

Killusions:
All the other routes like /mcp, /register, ... are served statically at root, making the only path it can be deployed on the root or with a proxy that strips the path in front (which we have). /callback for some reason is served at a dynamic route, which breaks with a proxy stripping the path. Either all paths should be dynamic or more easily all paths should be static


Killusions:
@zereight We were looking to use this new feature but it breaks with non-root paths.

zereight (approved):
Reviewed the current diff and focused callback-proxy behavior. No blocking issues found.

PR #466, opened 2026-05-11
feat: add create_group tool

Killusions:
Adds `create_group` tool - Create group/subgroup with name, path, description, visibility, and parent_id.

Also updates `list_namespaces` and `get_namespace` descriptions to clarify they cover groups (kind='group').


index.ts
```diff
@@ -8440,6 +8442,34 @@ async function handleToolCall(params: any) {
         };
       }
 
+      case "create_group": {
+        const args = CreateGroupSchema.parse(params.arguments);
+        const url = new URL(`${getEffectiveApiUrl()}/groups`);
```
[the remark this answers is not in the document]
Killusions: This doesn't make sense.

Killusions:
@zereight What do you think of this? It's currently missing compared to some implementations

zereight (changes requested):
The unresolved project-bound deployment issue still applies: when GITLAB_PROJECT_ID is set, repository-level mutators such as create_repository and fork_repository refuse to run, but create_group remains exposed and can POST /groups outside that configured project boundary. Please either guard create_group the same way in project-locked mode or explicitly remove it from the active tool surface for that mode, and add a small regression test for the behavior.

Killusions:
> The unresolved project-bound deployment issue still applies: when GITLAB_PROJECT_ID is set, repository-level mutators such as create_repository and fork_repository refuse to run, but create_group remains exposed and can POST /groups outside that configured project boundary. Please either guard create_group the same way in project-locked mode or explicitly remove it from the active tool surface for that mode, and add a small regression test for the behavior.

@zereight Ah I understand now, done and added a test (had to change the setup a bit)

test/test-geteffectiveprojectid.ts
```diff
@@ -57,17 +58,18 @@ describe('getEffectiveProjectId - No GITLAB_ALLOWED_PROJECT_IDS', () => {
       timeout: 5000,
       env: {
         STREAMABLE_HTTP: 'true',
+        REMOTE_AUTHORIZATION: 'true',
```
zereight, line 60: This test now starts the server with `REMOTE_AUTHORIZATION=true`, but the suite still connects with `StreamableHTTPTestClient`, which does not send any auth headers. When this file is run, the `before` hook fails with `Missing Private-Token, JOB-TOKEN, or Authorization header` before any project-id assertions execute.

Could we either keep these existing suites in server-managed token mode, or switch their client setup to `CustomHeaderClient({ authorization: `Bearer ${MOCK_TOKEN}` })`?

test/test-geteffectiveprojectid.ts
```diff
@@ -218,18 +220,17 @@ describe('getEffectiveProjectId - With multiple GITLAB_ALLOWED_PROJECT_IDS', ()
       port: mcpPort,
       timeout: 5000,
       env: {
-        STREAMABLE_HTTP: 'true',
+        REMOTE_AUTHORIZATION: 'true',
         GITLAB_API_URL: `${mockGitLabUrl}/api/v4`,
-        GITLAB_ALLOWED_PROJECT_IDS: `${DEFAULT_PROJECT_ID},${OTHER_PROJECT_ID}`,
-        GITLAB_READ_ONLY_MODE: 'true',
+        GITLAB_PROJECT_ID: DEFAULT_PROJECT_ID,
```
zereight: This suite is still named and asserted as the multiple-`GITLAB_ALLOWED_PROJECT_IDS` case, but it no longer configures `GITLAB_ALLOWED_PROJECT_IDS`; it sets only `GITLAB_PROJECT_ID`. After the auth-header issue is fixed, the `project_id: ''` test will resolve to the default project instead of throwing `Please specify a project ID`, so the suite no longer covers the multiple-allowed-project contract.

Could we restore `GITLAB_ALLOWED_PROJECT_IDS: `${DEFAULT_PROJECT_ID},${OTHER_PROJECT_ID}`` here, and omit `GITLAB_PROJECT_ID` if the intent is to require callers to choose explicitly?

test/test-geteffectiveprojectid.ts
```diff
@@ -276,15 +277,97 @@ describe('getEffectiveProjectId - With multiple GITLAB_ALLOWED_PROJECT_IDS', ()
...
+    try {
+      await client.callTool('fork_repository', { project_id: '999' });
+      assert.fail('Should have rejected fork_repository');
+    } catch (error) {
+      assert.ok(error instanceof Error);
+      assert.ok(error.message.includes('fork_repository is not allowed'), 'Should mention fork_repository');
+    }
+  });
+
+  test('should reject create_group when GITLAB_PROJECT_ID is set', async () => {
```
zereight, line 358: This is the important regression coverage for the project-locked `create_group` guard, but `test/test-geteffectiveprojectid.ts` is not included in `npm run test:mock` / `npm test`, so CI will not run it. That means this PR can regress the project-bound security check without a failing check.

Could we either add this file to the mock test script or move this regression into an existing test file that is already wired into CI?

zereight (changes requested):
Thanks for adding the project-lock guard; the runtime fix looks correct. I found remaining test coverage issues that keep this from being safe to merge yet.

zereight (changes requested):
🙏

test/test-geteffectiveprojectid.ts
Killusions: Done.

test/test-geteffectiveprojectid.ts
Killusions: Done.

test/test-geteffectiveprojectid.ts
Killusions: Done.

index.ts
```diff
@@ -8499,6 +8501,37 @@ async function handleToolCall(params: any) {
         };
       }
 
+      case "create_group": {
+        if (GITLAB_PROJECT_ID) {
+          throw new Error("Direct project ID is set. So create_group is not allowed");
+        }
+        const args = CreateGroupSchema.parse(params.arguments);
```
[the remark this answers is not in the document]
Killusions: Also done in a second commit.

Killusions:
@zereight Thank you, back to you

zereight (changes requested):
🙏

Killusions:
@zereight removed the readonly env and rebased

index.ts
```diff
@@ -8503,6 +8510,37 @@ async function handleToolCall(params: any) {
         };
       }
```
zereight, line 8518: With `GITLAB_ALLOWED_PROJECT_IDS` set (and no `GITLAB_PROJECT_ID`), `create_group` still reaches the global `POST /groups` endpoint outside the project boundary. The current guard only checks `GITLAB_PROJECT_ID`, so project-scoped deployments using the allowlist alone can create top-level groups regardless of the configured scope.

zereight (commented):
Thanks for addressing the `GITLAB_PROJECT_ID` case. I think we still need to handle `GITLAB_ALLOWED_PROJECT_IDS` here as well.

These two settings have slightly different roles: `GITLAB_PROJECT_ID` is primarily a default project fallback when callers omit `project_id`, while `GITLAB_ALLOWED_PROJECT_IDS` is an explicit project access boundary — a single value acts like a locked default project, and multiple values restrict access to that project set.

Since `create_group` targets the global `/groups` endpoint and is not scoped to any project, an allowlist-only deployment could still create groups outside the configured project boundary. Could we reject `create_group` whenever either project-scoped setting is active, e.g. `GITLAB_PROJECT_ID || GITLAB_ALLOWED_PROJECT_IDS.length > 0`, and add a small regression test for the allowlist-only case?

index.ts
Killusions: Done, also for fork, and create repository.

Killusions:
@zereight Back to you

index.ts
```diff
@@ -8510,6 +8517,37 @@ async function handleToolCall(params: any) {
...
 
+      case "create_group": {
```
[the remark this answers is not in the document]
Killusions: This is in line with the rest of the tools, the scope of this PR has ballooned enough.

zereight (approved):
Thanks!

Killusions:
@zereight thank you, could you maybe also create a release?

PR #487, opened 2026-05-23
fix(schemas): accept label objects in GitLabMergeRequestSchema response

### LINKED ISSUE #485: list_merge_requests: GitLabMergeRequestSchema rejects label objects in response
## Bug Description

When calling `list_merge_requests` (or any tool returning data parsed through `GitLabMergeRequestSchema`), Zod parsing fails on responses where GitLab returns labels as full label *objects* (with `id`, `name`, `color`, …) instead of plain strings.

Error:
```
0.labels.0: Expected string, received object
```

GitLab returns label objects on several endpoints — for example when the project has `with_labels_details` enabled by default at the instance, or when the underlying endpoint includes scoped/grouped label expansions. The current schema declares:

```ts
labels: z.array(z.string()).optional(),
```

so Zod refuses the entire response and the tool errors out.

## Steps to Reproduce

1. Call `list_merge_requests` against a project where GitLab returns labels as objects (e.g. internal GitLab instance with label details expansion).
2. Tool fails with `MCP error -32603: Invalid arguments` and Zod path `0.labels.0: Expected string, received object`.

## Expected Behavior

Parsing should accept both shapes that GitLab actually returns:
- `["bug", "enhancement"]` — array of label name strings
- `[{ id, name, color, text_color, description, … }]` — array of label objects (matches `GitLabLabelSchema`)

## Root Cause

`GitLabMergeRequestSchema.labels` (`schemas.ts` ~line 1200) only accepts `z.array(z.string())`. The sibling response schema `GitLabIssueSchema.labels` (`schemas.ts:1096`) already handles both shapes via:

```ts
labels: z.array(GitLabLabelSchema).or(z.array(z.string())), // Support both label objects and strings
```

The MR response schema should mirror this pattern.

Note: this is distinct from #380 / #431, which dealt with *input* tool schemas where LLMs send JSON-stringified arrays. `coerceStringArray` does not apply here because its inner schema is `z.array(z.string())`, which would still reject label objects from the API.

## Suggested Fix

In `schemas.ts`, change `GitLabMergeRequestSchema.labels` to mirror `GitLabIssueSchema`:

```ts
labels: z.array(GitLabLabelSchema).or(z.array(z.string())).optional(), // Support both label objects and strings
```

Plus regression tests in `test/schema-tests.ts` covering both shapes.

## Environment

- GitLab MCP version: latest (v2.1.14)
- Affected schema: `GitLabMergeRequestSchema`
- Affected tools: `list_merge_requests`, `get_merge_request`, and anything else parsing via `GitLabMergeRequestSchema`

Sid-96:
## Summary

- `list_merge_requests` / `get_merge_request` were failing with `0.labels.0: Expected string, received object` when GitLab returned labels as full label objects (with `id`, `name`, `color`, …) instead of plain strings.
- Mirror the pattern already used by `GitLabIssueSchema` (`schemas.ts:1096`) so the MR response schema accepts both shapes.
- Add regression tests in `runGitLabMergeRequestSchemaTests` covering string-array, label-object-array, and omitted labels.

Fixes #485

### Change

```diff
-  labels: z.array(z.string()).optional(),
+  labels: z.array(GitLabLabelSchema).or(z.array(z.string())).optional(), // Support both label objects and strings
```

## Test plan

- [x] `npm run test:schema` — 92 passed, 0 failed (3 new MR label cases included)
- [x] Manual: call `list_merge_requests` against a project whose GitLab instance returns labels as objects; confirm no Zod error and labels are preserved.

### DIFF
--- a/schemas.ts
+++ b/schemas.ts
@@ -1197,7 +1197,7 @@ export const GitLabMergeRequestSchema = z.object({
     .describe("Whether rebase is currently in progress for this merge request"),
   merge_when_pipeline_succeeds: z.coerce.boolean().optional(),
   squash: z.coerce.boolean().optional(),
-  labels: z.array(z.string()).optional(),
+  labels: z.array(GitLabLabelSchema).or(z.array(z.string())).optional(), // Support both label objects and strings
 });
 
 export const LineRangeSchema = z
--- a/test/schema-tests.ts
+++ b/test/schema-tests.ts
@@ -846,6 +846,56 @@ function runGitLabMergeRequestSchemaTests(): { passed: number; failed: number } 
       },
       validate: (data: Record<string, any>) => data.milestone === undefined,
     },
+    {
+      name: 'schema:gitlab_merge_request:labels-string-array',
+      input: {
+        ...baseMergeRequest,
+        labels: ['bug', 'enhancement'],
+      },
+      validate: (data: Record<string, any>) =>
+        Array.isArray(data.labels) &&
+        data.labels.length === 2 &&
+        data.labels[0] === 'bug' &&
+        data.labels[1] === 'enhancement',
+    },
+    {
+      name: 'schema:gitlab_merge_request:labels-object-array',
+      input: {
+        ...baseMergeRequest,
+        labels: [
+          {
+            id: 1,
+            name: 'bug',
+            color: '#ff0000',
+            text_color: '#ffffff',
+            description: null,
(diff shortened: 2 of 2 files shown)


zereight (approved):
Thanks — aligns GitLabMergeRequestSchema with GitLabIssueSchema for label objects vs strings. Regression tests wired in test:schema.

PR #489, opened 2026-05-23
feat(list-mrs): support approved_by_usernames filter

### LINKED ISSUE #488: list_merge_requests: support approved_by_usernames filter
## Description

`list_merge_requests` does not currently expose the GitLab API's `approved_by_usernames` query parameter. The field is not declared in `ListMergeRequestsSchema`, so even if a caller passes it, Zod strips the unknown key before the value ever reaches the HTTP request. As a result there is no way to filter merge requests by the usernames that approved them through this tool.

GitLab's `GET /merge_requests` (and the project-scoped variant) documents `approved_by_usernames` as a supported filter — see https://docs.gitlab.com/api/merge_requests/#list-merge-requests. It must be sent in bracket-array form (`approved_by_usernames[]=alice&approved_by_usernames[]=bob`), which the current query-string builder in `listMergeRequests` does not produce — it only special-cases `labels` (joined with `,`) and otherwise falls through to `String(value)`, which would serialize an array as `alice,bob` and silently mismatch.

## Expected Behavior

Callers can pass:

```json
{ "approved_by_usernames": ["alice", "bob"] }
```

to `list_merge_requests` and receive only MRs approved by any of the listed usernames. The request to GitLab carries the bracket-array form:

```
?approved_by_usernames[]=alice&approved_by_usernames[]=bob
```

LLM clients passing a JSON-stringified array (`'["alice","bob"]'`) should also work, consistent with how the `labels` field is coerced via `coerceStringArray` in the same schema.

## Actual Behavior

The argument is silently dropped at schema parse time; response is identical to a call without the filter.

## Suggested Fix

Two edits:

1. `schemas.ts` — declare the field on `ListMergeRequestsSchema` using the existing `coerceStringArray` helper so the value isn't stripped and LLM-stringified arrays are accepted:

   ```ts
   approved_by_usernames: coerceStringArray
     .optional()
     .describe("Returns merge requests approved by the given usernames (array)."),
   ```

2. `index.ts` — extend the query-string builder in `listMergeRequests` to forward this key in bracket-array form (mirroring how GitLab expects array filters):

   ```ts
   else if (key === "approved_by_usernames" && Array.isArray(value)) {
     for (const v of value) {
       url.searchParams.append(`${key}[]`, String(v));
     }
   }
   ```

## Environment

- GitLab MCP version: latest (v2.1.14)
- Affected schema: `ListMergeRequestsSchema`
- Affected tool: `list_merge_requests`

Sid-96:
## Summary

- `list_merge_requests` did not declare `approved_by_usernames` on `ListMergeRequestsSchema`, so Zod silently stripped the value before it reached the HTTP layer — the filter had no effect.
- Even if the value made it through, the existing query-string builder in `listMergeRequests` only special-cased `labels` and would have serialized an array as `alice,bob` via `String(value)`, which GitLab does not accept for this filter (it expects bracket-array form).

Fixes #488

### Changes

1. `schemas.ts` — declare the field using the existing `coerceStringArray` helper so the value is preserved and LLM-sent stringified arrays are coerced (consistent with how `labels` is handled on the same schema):

   ```ts
   approved_by_usernames: coerceStringArray
     .optional()
     .describe("Returns merge requests approved by the given usernames (array).")
   ```

2. `index.ts` — forward in bracket-array form:

   ```ts
   else if (key === "approved_by_usernames" && Array.isArray(value)) {
     for (const v of value) {
       url.searchParams.append(`${key}[]`, String(v));
     }
   }
   ```

### Tests

- `test/schema-tests.ts` — new `runApprovedByUsernamesSchemaTests` covering native array, JSON-stringified array, and omitted field. (`npm run test:schema` → 92 passed, 0 failed.)
- `test/test-list-merge-requests.ts` — new integration test that registers a custom mock handler to capture the outgoing request URL and asserts it contains `approved_by_usernames[]=alice&approved_by_usernames[]=bob` (URL-encoded). (4/4 passing.)

## Test plan

- [x] `npm run test:schema` passes
- [x] `npx tsx test/test-list-merge-requests.ts` passes, including the bracket-form assertion
- [x] Manual: call `list_merge_requests` against a real GitLab instance with `{ "approved_by_usernames": ["alice"] }` and confirm only MRs approved by `alice` are returned.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1933,6 +1933,11 @@ async function listMergeRequests(
       if (key === "labels" && Array.isArray(value)) {
         // Handle array of labels
         url.searchParams.append(key, value.join(","));
+      } else if (key === "approved_by_usernames" && Array.isArray(value)) {
+        // GitLab expects array-bracket form: approved_by_usernames[]=alice&approved_by_usernames[]=bob
+        for (const v of value) {
+          url.searchParams.append(`${key}[]`, String(v));
+        }
       } else {
         url.searchParams.append(key, String(value));
       }
--- a/schemas.ts
+++ b/schemas.ts
@@ -2037,6 +2037,9 @@ export const ListMergeRequestsSchema = z
       .string()
       .optional()
       .describe("Returns merge requests which have the user as a reviewer by username. Mutually exclusive with reviewer_id."),
+    approved_by_usernames: coerceStringArray
+      .optional()
+      .describe("Returns merge requests approved by the given usernames (array)."),
     created_after: z
       .string()
       .optional()
--- a/test/schema-tests.ts
+++ b/test/schema-tests.ts
@@ -1104,6 +1104,79 @@ function runLabelsCoercionSchemaTests(): { passed: number; failed: number } {
   return { passed, failed };
 }
 
+function runApprovedByUsernamesSchemaTests(): { passed: number; failed: number } {
+  console.log('\n=== ListMergeRequestsSchema approved_by_usernames Tests ===');
+
+  const cases: {
+    name: string;
+    input: Record<string, any>;
+    expected?: string[];
+    shouldFail?: boolean;
+  }[] = [
+    {
+      name: 'schema:list_merge_requests:approved_by_usernames-native-array',
+      input: { approved_by_usernames: ['alice', 'bob'] },
+      expected: ['alice', 'bob'],
+    },
+    {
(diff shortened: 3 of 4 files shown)


zereight (approved):
LGTM — preserves approved_by_usernames through ListMergeRequestsSchema and forwards bracket-array query params. Schema + integration tests verified.

PR #492, opened 2026-05-26
fix: group-level work items returning 404

vrajpal-jhala:
## Problem

Work items scoped to a GitLab **group** (URLs of the form `/groups/<path>/-/work_items/<iid>`) could not be fetched or manipulated via any work item tool. The `project_id` parameter was resolved through the REST `/projects/:id` endpoint, which returns 404 for group paths — causing the request to fail before even reaching the GraphQL layer.

## Root Cause

`resolveProjectPath()` called `GET /projects/:id` to obtain `path_with_namespace`, then passed the result to GraphQL's `namespace(fullPath: ...)`. That REST endpoint only resolves projects, not groups. No fallback existed for the 404 case.

Additionally, two functions — `resolveWorkItemGID` and `listIssueStatuses` — had their own **inline** `GET /projects/:id` fetches that completely bypassed `resolveProjectPath`, so they failed for group paths even after the fallback was added.

GitLab's GraphQL `namespace(fullPath: ...)` field natively accepts both project paths and group paths, so the REST lookup was an unnecessary blocker for group-level namespaces.

### `index.ts` — `resolveProjectPath`

Added a two-step lookup: when `GET /projects/:id` returns 404, the function now falls back to `GET /groups/:id`. If that succeeds, it returns `group.full_path` for downstream GraphQL use. If the group lookup also returns a non-ok response (404, 429, 5xx, etc.), that error is surfaced directly via `handleGitLabError` — preserving fail-fast behaviour and preventing real errors (rate limits, server faults) from being masked by the original project 404.

### `index.ts` — `resolveWorkItemGID` and `listIssueStatuses`

Both functions previously had their own inline `GET /projects/:id` REST fetch for path resolution, bypassing `resolveProjectPath` entirely. Replaced both with a direct call to `resolveProjectPath(projectId)`, making all work item GraphQL functions route through the single group-aware resolver.

**All 8 `resolveProjectPath` call sites are now work-item functions only.** No REST-based tool (issues, MRs, branches, etc.) was changed — those correctly remain project-scoped.

### `schemas.ts` — all work item schemas

Added a shared `ProjectIdOrPathSchema` constant and applied it to the `project_id` field across all 16 work item tool schemas (`WorkItemParamsSchema`, `ListWorkItemsSchema`, `CreateWorkItemSchema`, `ConvertWorkItemTypeSchema`, `ListWorkItemStatusesSchema`, `ListWorkItemNotesSchema`, `CreateWorkItemNoteSchema`, `MoveWorkItemSchema`, `ListCustomFieldDefinitionsSchema`, and all 7 emoji reaction / timeline event schemas). The description now explicitly states that a group path (e.g. `group/subgroup`) is a valid value, so AI clients pass it rather than failing to find a project ID.

## Behaviour

| Input | Before | After |
|---|---|---|
| Numeric project ID | ✅ Works | ✅ Works |
| Project path (`group/project`) | ✅ Works | ✅ Works |
| Group path (`group/subgroup`) | ❌ 404 thrown | ✅ Resolved via `GET /groups/:id`, passed to GraphQL |
| Typo / deleted / permission 404 | ❌ 404 thrown | ❌ 404 thrown (fail-fast preserved) |
| Rate limit (429) on group lookup | ❌ Masked as project 404 | ❌ 429 surfaced directly |
| Server error (5xx) on group lookup | ❌ Masked as project 404 | ❌ 5xx surfaced directly |

## Notes

- Existing project-level work item flows are unaffected — REST still resolves `path_with_namespace` on the first call, with no overhead.
- The group fallback only incurs an extra REST round-trip on a project 404, so there is no performance impact for the common case.
- No new tools or parameters added; the fix works with the existing `project_id` field.
- REST-based tools (issues, MRs, branches, labels, wikis, etc.) are intentionally unchanged — the GitLab REST API's `/projects/:id` endpoint only resolves projects, so advertising group path support there would be incorrect.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -2894,10 +2894,21 @@ async function resolveProjectPath(projectId: string): Promise<string> {
   const projectResponse = await fetch(projectUrl.toString(), {
     ...getFetchConfig(),
   });
-  // If the path resolves to a group rather than a project (404), use it directly —
-  // the GraphQL `namespace(fullPath: ...)` field accepts both project and group paths.
+  // If the project lookup returned 404, check whether the path is a real group
+  // before falling through to GraphQL. This prevents typos, deleted projects, and
+  // permission-scoped 404s from silently producing empty GraphQL results.
   if (projectResponse.status === 404) {
-    return effectiveProjectId;
+    const groupUrl = new URL(
+      `${getEffectiveApiUrl()}/groups/${encodeURIComponent(effectiveProjectId)}`
+    );
+    const groupResponse = await fetch(groupUrl.toString(), {
+      ...getFetchConfig(),
+    });
+    if (groupResponse.ok) {
+      const group: any = await groupResponse.json();
+      return group.full_path as string;
+    }
+    // Not a group either — fall through to the outer error handler below
   }
   await handleGitLabError(projectResponse);
   const project: any = await projectResponse.json();
--- a/index.ts
+++ b/index.ts
@@ -2894,6 +2894,11 @@ async function resolveProjectPath(projectId: string): Promise<string> {
   const projectResponse = await fetch(projectUrl.toString(), {
     ...getFetchConfig(),
   });
+  // If the path resolves to a group rather than a project (404), use it directly —
+  // the GraphQL `namespace(fullPath: ...)` field accepts both project and group paths.
+  if (projectResponse.status === 404) {
+    return effectiveProjectId;
+  }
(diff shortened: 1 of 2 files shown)


zereight (approved):
LGTM. Group path fallback in resolveProjectPath with confirmed GET /groups lookup, numeric ID gate, and unified work-item path resolution. Addresses prior Codex P1/P2 feedback.

PR #495, opened 2026-05-30
fix: defer OAuth token acquisition to first use; stop browser popups on transient refresh errors

medyas:
## Problem

When `GITLAB_USE_OAUTH=true`, the server acquires the OAuth token **eagerly and synchronously at startup** (`runServer` → `initializeOAuthClient` → `getAccessToken`), blocking the MCP handshake on a network call to GitLab.

For a self-hosted GitLab (often behind a VPN), the network frequently isn't ready at the instant the MCP client spawns the server. The refresh `fetch` then hangs past the client's connection timeout, and on *any* refresh error the code falls through to the interactive `startOAuthFlow()` — **opening a browser on every cold start**. Reproduced reliably: the client reports `connection timed out after 30000ms` while the server is still blocked logging `Token expired. Refreshing...`, then a browser window opens.

Two distinct issues:
1. Auth is eager + blocking at startup, even though a lazy path (`ensureValidOAuthToken`, per tool call) already exists.
2. The refresh `catch` treats every failure — including transient network/timeout errors — as "must re-authenticate interactively", so a network blip opens a browser.

## Fix

- **Defer token acquisition.** Construct the OAuth client synchronously at startup (no network) via a new `createGitLabOAuthClient`. The token is acquired lazily on the first tool call, by which point the network is ready. `initializeOAuthClient` is kept and now delegates to the new constructor.
- **Fast-fail refresh.** Add a 10s `AbortSignal.timeout` to the refresh request so it fails fast instead of hanging.
- **Only re-auth interactively when warranted.** New pure helper `isAuthInvalidTokenResponse(status, body)` returns true only for a 400/401 with `invalid_grant`/`invalid_token`. The refresh `catch` opens the browser only in that case; transient and configuration errors propagate without a popup.

## Tests

- New `test/oauth-refresh-classify.test.ts` — 8 cases for the classifier (invalid_grant/invalid_token → true; invalid_client/invalid_request/5xx/429/non-JSON → false), wired into `test:mock`.
- Existing `test/test-auth-retry.ts` still green (19/19). `npm run build` clean.


zereight (changes requested):
Thanks for the fix — deferring OAuth startup and only opening the browser for genuinely invalid refresh tokens is the right direction.

I agree with the open Codex comment about populating `OAUTH_ACCESS_TOKEN` when a valid cached token already exists. Please address that before merge.

A few additional items:

1. **Regression test** — after the cached-token fix, please add a small test that covers the first tool call with a valid on-disk token (assert `buildAuthHeaders()` / the outgoing request includes `Authorization`). The new `isAuthInvalidTokenResponse` unit tests are helpful, but they do not exercise the lazy startup path end-to-end.

2. **Rebase onto current `main`** — this branch will conflict in `index.ts` and `package.json` after recent merges.

Happy to take another look once the cached-token path is fixed and tests are in place.

nick-strause-34485:
@medyas hey, do you plan on carrying this forward?

PR #505, opened 2026-06-07
feat: `create_repository` should support `namespace_id`

andrea-veritas:
### Problem

The `create_repository` tool currently creates projects only in the user's personal namespace. There is no way to create a repository under a specific Group, which is a common requirement for team workflows (e.g., creating a documentation repo under a project Group).

The [GitLab REST API `POST /projects`](https://docs.gitlab.com/api/projects/#create-a-project) natively supports the `namespace_id` parameter to specify the target namespace. However, the MCP tool's schema does not expose this parameter.

### Steps to Reproduce

```typescript
// Current behavior — project always ends up in user namespace
mcp_gitlab_create_repository({
  name: "my-docs",
  namespace_id: 6,  // silently ignored
  visibility: "private",
  initialize_with_readme: true
})
// Result: path_with_namespace = "gs/my-docs" (user namespace)
// Expected: path_with_namespace = "my-group/my-docs" (Group namespace)
```

### Proposed Solution

Add an optional `namespace_id` parameter to `CreateRepositoryOptionsSchema` and `CreateRepositorySchema` in `schemas.ts`, and pass it through in the `createRepository` handler in `index.ts`.

**schemas.ts:**

```diff
 export const CreateRepositoryOptionsSchema = z.object({
   name: z.string().describe("Repository name"),
+  namespace_id: z.coerce.number().int().optional().describe("Namespace group ID"),
   description: z.string().optional(),
   // ...
 });
```

**index.ts — createRepository function:**

```diff
   body: JSON.stringify({
     name: options.name,
+    ...(options.namespace_id !== undefined ? { namespace_id: options.namespace_id } : {}),
     description: options.description,
     // ...
   }),
```

### Additional Context

This change is backward-compatible — when `namespace_id` is not provided, behavior is unchanged. I have tested this fix against a self-hosted GitLab CE 19.0.1 instance and confirmed it works correctly. A minimal implementation (2 source files, 3 lines added) is available in the `fix/create_project_with_namespace_id` branch of [my fork](https://github.com/andrea-veritas/gitlab-mcp/tree/fix/create_project_with_namespace_id).

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -235,6 +235,7 @@ import {
   CreateProjectMilestoneSchema,
   CreateRepositoryOptionsSchema,
   CreateRepositorySchema,
+  AddGroupMemberSchema,
   CreateGroupSchema,
   CreateWikiPageSchema,
   CreateGroupWikiPageSchema,
@@ -9068,6 +9069,32 @@ async function handleToolCall(params: any) {
 
         return {
           content: [{ type: "text", text: JSON.stringify(group, null, 2) }],
+        };
+      }
+
+      case "add_group_member": {
+        rejectIfProjectScopedDeployment("add_group_member");
+        const gmArgs = AddGroupMemberSchema.parse(params.arguments);
+        const { group_id, user_id, access_level, expires_at } = gmArgs;
+        const gmUrl = `${getEffectiveApiUrl()}/groups/${group_id}/members`;
+
+        const gmBody: Record<string, unknown> = {
+          user_id,
+          access_level,
+        };
+        if (expires_at) gmBody.expires_at = expires_at;
+
+        const gmResponse = await fetch(gmUrl, {
+          method: "POST",
+          headers: { ...getFetchConfig().headers, "Content-Type": "application/json" },
+          body: JSON.stringify(gmBody),
+        });
+
+        await handleGitLabError(gmResponse);
+        const gmData = await gmResponse.json();
+
+        return {
+          content: [{ type: "text", text: JSON.stringify(gmData, null, 2) }],
         };
       }
 
--- a/index.ts
+++ b/index.ts
@@ -4689,6 +4689,7 @@ async function createRepository(
     method: "POST",
     body: JSON.stringify({
       name: options.name,
+      ...(options.namespace_id !== undefined ? { namespace_id: options.namespace_id } : {}),
       description: options.description,
       visibility: options.visibility,
       initialize_with_readme: options.initialize_with_readme,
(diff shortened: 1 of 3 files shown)


morsik:
Absolutely would love this feature!

I just realized this wasn't a thing when used Sonnet to lower token cost... apparently Opus cheated a little and ran HTTP request to Gitlab API directly automatically under the hood instead, while Sonnet informed me about missing feature and stop his work.

zereight (changes requested):
Thanks for the contribution — the `namespace_id` support for `create_repository` looks correct and backward compatible.

Before merge, could you please:

1. **Split the PR scope** — this branch also adds a full `add_group_member` tool (second commit). Please keep #505 focused on `namespace_id` and open a separate PR for group membership.
2. **Rebase onto current `main`** — the branch is behind recent merges and will conflict in `index.ts`.
3. **Add a small regression test** — assert that `namespace_id` is forwarded in the `POST /projects` body when provided.

For the follow-up `add_group_member` PR (not blocking this one once split):
- Prefer `group_id: z.coerce.string()` to match other group tools and support path-based IDs.
- Restrict `access_level` to GitLab’s allowed values (10/20/30/40/50).
- Update `test/test-toolset-filtering.ts` for the expanded `groups` toolset.

Happy to take another look once the scope is narrowed and tests are in place.

index.ts
```diff
@@ -9070,6 +9072,32 @@ async function handleToolCall(params: any) {
         };
       }
 
+      case "add_group_member": {
```
zereight, line 9075: This looks like a separate feature from `namespace_id`. Could you move `add_group_member` to its own PR so this one stays focused and easier to review?

schemas.ts
```diff
@@ -4192,3 +4194,10 @@ export const ListDependencyProxyBlobsSchema = z.object({
 export const PurgeDependencyProxyCacheSchema = z.object({
   group_id: z.coerce.string().describe("Group ID or URL-encoded path"),
 });
+
+export const AddGroupMemberSchema = z.object({
+  group_id: z.coerce.number().int().min(1).describe("GitLab Group ID"),
```
zereight, line 4199: Most group tools in this repo use `z.coerce.string()` for `group_id` (path or numeric ID). Consider aligning here in the follow-up PR.

zereight:
Thanks again for opening this — the `namespace_id` support was exactly what we needed.

I ended up landing that part in #530 (rebased on current `main`, with a small regression test), so I'm closing this PR to avoid duplicate work.

If you still want to contribute `add_group_member`, a separate focused PR would be great — happy to review when you have time.

PR #508, opened 2026-06-11
fix: derive remote download URLs from forwarded headers

pierredeman:
## Summary

- derive remote download URLs from `Forwarded` / `X-Forwarded-*` headers when `MCP_SERVER_URL` is unset
- preserve the derived public base URL in remote auth session context
- cover `download_job_artifacts` in Streamable HTTP remote auth mode

## Why

Remote deployments behind reverse proxies can run with `HOST=0.0.0.0`. When `MCP_SERVER_URL` is not configured, remote download tools can return URLs like `http://0.0.0.0:3002/downloads/...`, which are not reachable by clients.

Reverse proxies commonly provide the public origin through forwarded headers, so this uses those headers as a fallback while keeping `MCP_SERVER_URL` authoritative.

## Tests

- `npm run build`
- `node --import tsx/esm --test test/test-remote-downloads.ts`
- tested behind Traefik in a temporary Portainer stack without `MCP_SERVER_URL`


zereight (changes requested):
Thanks for tackling the reverse-proxy download URL issue — the overall approach (session `publicBaseUrl` + `buildDownloadUrl` fallback chain) looks sound.

A few items before merge (complementing the open CodeRabbit threads):

1. **Wire the new tests into CI** — `test/test-remote-downloads.ts` is not listed in `package.json` `test:mock`, so the added regression coverage will not run in PR CI. Please add it to the mock test script.

2. **Trust-proxy gating needs a REMOTE_AUTHORIZATION path** — gating forwarded-header parsing on `req.app.get("trust proxy")` is the right idea, but today `app.set("trust proxy", 1)` is only enabled when `GITLAB_MCP_OAUTH=true`. This PR targets `REMOTE_AUTHORIZATION` deployments behind a proxy, so the gate would stay off unless you also enable trust proxy for Streamable HTTP / remote-auth mode (or add an explicit env flag such as `TRUST_PROXY`).

3. **Document the fallback** — please note in `docs/environment-variables.md` (and briefly in README) that when `MCP_SERVER_URL` is unset, download URLs may be derived from `Forwarded` / `X-Forwarded-*`, and that this should only be relied on behind a trusted reverse proxy.

4. **Rebase onto current `main`** — `index.ts` will conflict after recent merges.

Happy to take another look once these are addressed along with the remaining CodeRabbit items.

pierredeman:
Thanks for the detailed review.

I rebased on the latest `main` and updated the implementation to gate forwarded-header URL derivation behind `MCP_TRUST_PROXY`, keep `MCP_SERVER_URL` as the explicit override, wire the remote download tests into `test:mock`, and harden forwarded host/prefix handling.

Following the CodeRabbit feedback, `GITLAB_MCP_OAUTH` no longer enables Express `trust proxy` implicitly. This is a behavior change for existing OAuth deployments behind a reverse proxy: they should now set `MCP_TRUST_PROXY=true` explicitly if they rely on forwarded headers / real client IP handling.

This should now match the trust-proxy direction you suggested while keeping the behavior opt-in.

zereight:
Thanks for addressing the review feedback — `MCP_TRUST_PROXY` gating and the added tests look good from my side.

Before merge, please rebase onto current `main`. The branch is behind (including recent `package.json` / lockfile changes), and I expect a conflict in `package.json`. Push after rebasing so CI can run on the updated head.

zereight (approved):
LGTM. The trust-proxy gate, forwarded-header hardening, docs, and remote download tests all look good now.

vrajpal-jhala:
Hi, @zereight @pierredeman. This PR introduced a silent breaking change for OAuth deployments behind a reverse proxy. `trust proxy` was previously unconditionally set in OAuth mode; it's now opt-in via `MCP_TRUST_PROXY=true`. Existing deployments will see `ERR_ERL_UNEXPECTED_X_FORWARDED_FOR` from `express-rate-limit` on every OAuth endpoint request, causing auth handshake failures, connection drops, and tool call timeouts. Worth adding a migration note to the `v2.1.22` release.

Also: `startSSEServer` was not updated to respect `MCP_TRUST_PROXY` - only `startStreamableHTTPServer` got the fix.

> [!WARNING]
> If you were running OAuth mode behind a reverse proxy before this update and haven't set `MCP_TRUST_PROXY=true` yet, add it to your deployment - the old implicit trust proxy behaviour is gone.

pierredeman:
Thanks for calling this out.

You’re right about the migration impact, as i said here : https://github.com/zereight/gitlab-mcp/pull/508#issuecomment-4682062764
I agree this should be added clearly in the `v2.1.22` release notes.

Good catch on `startSSEServer` as well.  
OAuth is Streamable HTTP only, so the OAuth breakage does not apply to SSE, but `MCP_TRUST_PROXY` is a transport-level setting and SSE should honor it consistently.
I opened a small PR for that : https://github.com/zereight/gitlab-mcp/pull/518

vrajpal-jhala:
> Thanks for calling this out.
> 
> You’re right about the migration impact, as i said here : [#508 (comment)](https://github.com/zereight/gitlab-mcp/pull/508#issuecomment-4682062764) I agree this should be added clearly in the `v2.1.22` release notes.
> 
> Good catch on `startSSEServer` as well. OAuth is Streamable HTTP only, so the OAuth breakage does not apply to SSE, but `MCP_TRUST_PROXY` is a transport-level setting and SSE should honor it consistently. I opened a small PR for that : #518

Looks good @pierredeman. I was facing another issue where nginx was setting header with ip + port which was breaking the rate limit config. I have applied a fix in a separate #519.

PR #513, opened 2026-06-14
feat: binary (base64) commits and per-file actions in push_files

hodyhq:
## What & why

`push_files` currently hardcodes `action: "create"` for every file, so it:

- errors with *"A file with this name already exists"* on any existing file (can't update),
- can't delete or move files, and
- treats all content as text, so committing **binary** files (images, etc.) is impossible.

I hit all three building a small project with this MCP (committing PNG logos and updating files in one commit).

## Changes

- **`push_files`** — each file entry gains optional fields:
  - `action`: `create` | `update` | `delete` | `move` (default `create`)
  - `encoding`: `text` | `base64` — base64 content is passed through untouched (binary-safe)
  - `previous_path` (for `move`); `content` is now optional (omit for `delete`)
- **`create_or_update_file`** — gains optional `encoding: "base64"` for binary files.

Defaults preserve existing behaviour — the global `GITLAB_REPO_FILE_ENCODING` toggle still works, and text commits are unchanged. The base64 path sends content straight to GitLab's commits/files API with `encoding: "base64"` instead of double-encoding it.

## Tests

Added `PushFilesSchema` / `CreateOrUpdateFileSchema` cases to `test/schema-tests.ts` (action / encoding / optional-content / move, plus rejection of invalid `action`/`encoding`). `npm run build` is clean and the schema suite reports **109 passed, 0 failed**.

Also verified end-to-end against a live GitLab: a PNG committed via `push_files` (`encoding: "base64"`) round-trips byte-for-byte (sha256 match), and `update` / `delete` actions work as expected.

### DIFF
--- a/activate-fork.sh
+++ b/activate-fork.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+# Point Claude Code's `gitlab` MCP server at the hodyhq fork (binary + update support).
+# RUN THIS WITH CLAUDE CODE FULLY QUIT, then relaunch — the running app owns
+# ~/.claude.json and will clobber the edit otherwise.
+#
+#   bash ~/.dev/Projects/gitlab-mcp/activate-fork.sh          # activate fork
+#   bash ~/.dev/Projects/gitlab-mcp/activate-fork.sh --revert # back to upstream npx
+set -euo pipefail
+
+CFG="$HOME/.claude.json"
+BUILD="/home/hody/.dev/Projects/gitlab-mcp/build/index.js"
+MODE="${1:-activate}"
+
+[ -f "$CFG" ] || { echo "no $CFG"; exit 1; }
+[ "$MODE" = "activate" ] && [ ! -f "$BUILD" ] && { echo "build missing: $BUILD (run: cd ~/.dev/Projects/gitlab-mcp && npm install && npm run build)"; exit 1; }
+
+cp "$CFG" "$CFG.bak.$(date +%s)"
+
+python3 - "$CFG" "$BUILD" "$MODE" <<'PY'
+import json, sys
+cfg, build, mode = sys.argv[1], sys.argv[2], sys.argv[3]
+d = json.load(open(cfg)); n = 0
+def walk(o):
+    global n
+    if isinstance(o, dict):
+        ms = o.get("mcpServers")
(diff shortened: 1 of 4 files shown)


hodyhq:
Thanks for the review! Addressed in the latest commit:

- **`encoding: "text"` ignored when global encoding is base64** — fixed. An explicit per-file `encoding` (text *or* base64) now always takes precedence; the global `GITLAB_REPO_FILE_ENCODING` toggle is used only when `encoding` is omitted.
- **`move` blanking content** — fixed. For a `move` with no `content`, the content field is now omitted entirely so GitLab preserves the original file's content (instead of sending `""`).
- Added schema tests for explicit `text` encoding and `move` without `previous_path`.

Schema suite: **111 passed / 0 failed**, build clean.

zereight (changes requested):
Thanks for the push_files work — per-file actions and base64 support are a real gap today, and the follow-up commit addressed the encoding/move issues from the first review round.

Before merge, a few things still need attention:

**`activate-fork.sh`** — This looks like a local fork-testing helper (hardcoded `/home/hody/...` path, Claude Code config editing). It shouldn't ship in the upstream repo. Please drop it from the PR.

**Rebase onto current `main`** — `index.ts` and `schemas.ts` both conflict with recent changes on main. Please rebase and resolve.

**Empty content on create/update** — The move-without-content case is handled correctly now, but create/update still coerces missing `content` to `""`, which can silently write an empty file. Either reject that at schema/runtime or document it clearly.

**Docs** — `docs/tools/repositories.md` still shows the old `push_files` shape. Run `make tools-docs` after the schema changes and commit the regenerated pages.

**Tests** — The new schema cases in `test/schema-tests.ts` are good, but there's no coverage for the actual `createCommit` payload mapping (base64 passthrough, explicit `encoding: "text"` vs global base64, delete/move omitting content). A small unit test around that mapping would help.

Happy to take another look once these are sorted.

zereight:
Thanks for the original work here — per-file actions and base64 were a real gap.

This has been stale since the June 17 review (rebase, drop activate-fork.sh, empty create/update content, docs, mapping tests). I landed the same feature on current main as #670 so we can finish it without the fork helper and with the review notes addressed.

Closing this one in favor of #670. Credit is already in the changelog.

PR #518, opened 2026-06-16
Honor MCP_TRUST_PROXY in SSE transport

pierredeman:
Follow-up to #508.

This applies the existing `MCP_TRUST_PROXY` behavior to the legacy SSE transport as well, so both HTTP transports use the same explicit trust-proxy toggle.

It also updates the environment variable docs to describe that `MCP_TRUST_PROXY` enables Express `trust proxy` for HTTP transports.


zereight:
Thanks for the follow-up on #508 — the SSE `MCP_TRUST_PROXY` change and the docs clarification were both needed.

We ended up landing the same SSE trust-proxy behavior in #519, and main now also has broader `MCP_TRUST_PROXY` docs/tests from a follow-up commit. So the functional part of this PR is already covered on `main`.

Closing this one to avoid merge conflicts on `README.md`, `index.ts`, and the moved docs path. If you still want the OAuth env table row or a CHANGELOG note from this branch, happy to cherry-pick those separately — just ping me.

zereight:
Superseded by #519 and follow-up docs on main.

PR #520, opened 2026-06-16
chore(docker): add .dockerignore file

pstoeckle:
I added a [`.dockerignore` file](https://docs.docker.com/build/concepts/context/#dockerignore-files) to

1. reduce the size of the build context,
2. keep files like `test-note.js` out of the final container image

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,3 +12,5 @@
 !.npmrc
 # Ignore test files
 test*.ts
+test
+mcp-server/test
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+# Ignore all files by default
+*
+# Include specific files and directories
+!*.ts
+!mcp-server
+!package-lock.json
+!package.json
+!stateless
+!tools
+!tsconfig.json
+!utils
+!.npmrc
+# Ignore test files
+test*.ts


zereight:
Thanks for adding `.dockerignore` — shrinking the Docker build context is worthwhile.

Could you take another look at the CodeRabbit feedback on this PR? In particular, the `test/` directory still looks like it would be included because `!*.ts` re-includes nested `.ts` files while `test*.ts` only covers root-level names. Please also double-check the directory allowlist patterns (`!utils/**`, etc.) and confirm `docker build` still passes with the proposed file.

Happy to re-review once that is addressed.

zereight (changes requested):
Please address the CodeRabbit thread on `.dockerignore` before merge.

Main concern: `test/` is not excluded, so `test/**/*.ts` can still enter the build context via `!*.ts` and get compiled during `prepare` → `tsc`. Add an explicit `test/` ignore (and consider `!dir/**` for allowed source dirs). Please verify `docker build` with the updated file.

pstoeckle:
> Please address the CodeRabbit thread on `.dockerignore` before merge.
> 
> Main concern: `test/` is not excluded, so `test/**/*.ts` can still enter the build context via `!*.ts` and get compiled during `prepare` → `tsc`. Add an explicit `test/` ignore (and consider `!dir/**` for allowed source dirs). Please verify `docker build` with the updated file.

Addressed in 33428eb

pstoeckle:
> Please verify `docker build` with the updated file.

I used `docker build --target builder . --tag test-ignore && docker run test-ignore CMD` and got the following results.

```console
$ ls -al /app
.npmrc
auth-retry.ts
config.ts
customSchemas.ts
gitlab-client-pool.ts
index.ts
mcp-server
oauth-proxy.ts
oauth.ts
package-lock.json
package.json
schemas.ts
stateless
tools
tsconfig.json
utils
```

and 

```console
$ ls -al /app/mcp-server
.gitignore
package-lock.json
package.json
src
tsconfig.json
```

Should be fine

zereight (approved):
Follow-up addressed the test/ exclusion concern. Simulated docker context build passes.

PR #568, opened 2026-07-03
fix: encode GitLab URL path IDs

zereight:
## Summary
- encode GitLab URL path ID segments before interpolation
- cover issue, merge request, note, discussion, link, draft note, version, and group ID path segments
- add regression coverage for path segment encoding

## Tests
- npm run build
- node --import tsx/esm --test test/path-segment-encoding.test.ts

---

> [!NOTE]
> **Medium Risk**
> Fixes path-segment injection / traversal in outbound GitLab API URLs across many call sites; behavior change is intentional encoding but wide surface area warrants careful review.
> 
> **Overview**
> **Hardens GitLab REST URL construction** so dynamic IDs in path segments are always percent-encoded instead of interpolated raw into template strings.
> 
> `encodeGitLabPathSegment` now accepts `unknown`, stringifies values, and uses decode-then-encode when possible, with a **fallback** to `encodeURIComponent` when percent-decoding throws (malformed sequences). That helper is applied across issue, merge request, discussion, note, link, draft note, version, and group search endpoints—anywhere `${issueIid}`, `${mergeRequestIid}`, `${discussionId}`, etc. previously appeared unencoded.
> 
> Adds **`test/path-segment-encoding.test.ts`**, which asserts those raw interpolation patterns are gone from `index.ts` and documents that encoded traversal-like payloads stay within a single path segment. The mock test suite runs this file first via `package.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69c7bc6a32481ff08122fd094e4ccb7693d467a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


zereight:
Addressed the valid review items in fb08054:

- removed redundant group_id pre-encoding before encodeGitLabPathSegment
- made encodeGitLabPathSegment tolerate malformed percent-encoded input
- kept the regression guard broad because these URL builders are private and this catches accidental raw segment interpolation across all affected paths.

@coderabbitai review

PR #598, opened 2026-07-14
fix: enforce GitLab path boundaries

ShiroKSH:
## Summary

- Enforce the configured project allowlist in `get_draft_note`.
- Reject dot path segments before GitLab URL construction, including artifact, release, discussion, note, and emoji endpoints.
- Prevent attachment filenames from influencing local output paths.
- Expire unauthenticated tool-discovery sessions so they release server capacity.

## Root cause

Several endpoints interpolated request identifiers directly into URL paths. Dot segments are not encoded by `encodeURIComponent`, so URL normalization could change the targeted endpoint. Public-discovery sessions were also created before a session ID existed, leaving their capacity slot without a timeout.

## Validation

- `npm run build`
- `node --import tsx/esm --test test/path-segment-encoding.test.ts`
- `node --import tsx/esm --test test/test-download-attachment.ts`
- `node --import tsx/esm --test --experimental-test-isolation=none test/streamable-http-unauthenticated-discovery.test.ts`
- `node --import tsx/esm --test --experimental-test-isolation=none test/test-geteffectiveprojectid.ts`
- `scripts/consumer-install-smoke.sh` via Git Bash

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1823,11 +1823,18 @@ function parseAllowedGitLabApiUrls(value: string): Array<{ host: string; apiUrl:
 
 function encodeGitLabPathSegment(value: unknown): string {
   const segment = String(value);
+  let decodedSegment: string;
   try {
-    return encodeURIComponent(decodeURIComponent(segment));
+    decodedSegment = decodeURIComponent(segment);
   } catch {
-    return encodeURIComponent(segment);
+    decodedSegment = segment;
   }
+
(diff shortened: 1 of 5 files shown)


zereight (approved):
LGTM. Path-boundary hardening looks solid: allowlist on get_draft_note, fail-closed . / .. segments, safe attachment filenames, and discovery session expiry. Local build + the listed path/attachment/discovery/allowlist tests passed.

PR #610, opened 2026-07-17
feat: add vulnerabilities toolset for AI-assisted triage

sidi7:
Add opt-in vulnerabilities toolset (GITLAB_TOOLSETS=vulnerabilities) with four tools: list_project_vulnerabilities, get_vulnerability, dismiss_vulnerability, and confirm_vulnerability.

- Zod input schemas in schemas.ts
- Tool registration, toolset definition, permission sets in tools/registry.ts
- API functions and handleToolCall cases in index.ts
- Skill reference at skills/gitlab-mcp/reference/vulnerability-triage.md
- Tests at test/test-vulnerabilities.ts (13 tests covering filters, request bodies, and readonly permission blocking)
- Generated docs at docs/tools/vulnerabilities.md

dismiss/confirm require modify or full permission mode (blocked in readonly). Neither is in deleteTools (available in modify mode as state transitions).

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -9239,6 +9239,9 @@ async function dismissVulnerability(
       `Failed to dismiss vulnerability: ${data.vulnerabilityDismiss.errors.join(", ")}`
     );
   }
+  if (!data.vulnerabilityDismiss.vulnerability) {
+    throw new Error(`Vulnerability not returned after dismissal (id: ${vulnerabilityId})`);
+  }
   return data.vulnerabilityDismiss.vulnerability;
 }
 
@@ -9272,6 +9275,9 @@ async function confirmVulnerability(
     throw new Error(
       `Failed to confirm vulnerability: ${data.vulnerabilityConfirm.errors.join(", ")}`
     );
+  }
+  if (!data.vulnerabilityConfirm.vulnerability) {
+    throw new Error(`Vulnerability not returned after confirmation (id: ${vulnerabilityId})`);
   }
   return data.vulnerabilityConfirm.vulnerability;
 }
--- a/index.ts
+++ b/index.ts
@@ -9030,11 +9030,17 @@ async function purgeDependencyProxyCache(groupId: string): Promise<void> {
 // params on list and reason/comment on dismiss. GitLab recommends the
 // GraphQL API for vulnerability management, which supports all of these.
 
-/** Convert a numeric vulnerability ID (or an existing GID) to a GraphQL global ID. */
+/** Convert a numeric vulnerability ID (or an existing Vulnerability GID) to a GraphQL global ID. */
 function toVulnerabilityGid(vulnerabilityId: string): string {
-  return vulnerabilityId.startsWith("gid://")
-    ? vulnerabilityId
-    : `gid://gitlab/Vulnerability/${vulnerabilityId}`;
+  if (/^\d+$/.test(vulnerabilityId)) {
+    return `gid://gitlab/Vulnerability/${vulnerabilityId}`;
+  }
+  if (/^gid:\/\/gitlab\/Vulnerability\/\d+$/.test(vulnerabilityId)) {
+    return vulnerabilityId;
+  }
+  throw new Error(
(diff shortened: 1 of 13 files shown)


index.ts
```diff
@@ -9020,6 +9024,264 @@ async function purgeDependencyProxyCache(groupId: string): Promise<void> {
...
+          fullPath
+        }
+      }
+    }`,
+    { id: toVulnerabilityGid(vulnerabilityId) }
+  );
+  if (!data.vulnerability) {
+    throw new Error(`Vulnerability not found: ${vulnerabilityId}`);
+  }
+  return data.vulnerability;
```
zereight, line 9271: `list_project_vulnerabilities` is allowlist-gated via `resolveProjectFullPath`, but get/dismiss/confirm only take `vulnerability_id` and hit GraphQL directly.

With `GITLAB_ALLOWED_PROJECT_IDS` set, callers can still touch vulns outside the allowed projects as long as the PAT can see them. We already block that pattern in `execute_graphql` / `rejectIfProjectScopedDeployment`, and we just closed the same hole on `get_draft_note` in #598.

Easiest fix: require `project_id`, or verify the vuln's project against the allowlist after fetch, before return/mutate. Please add a deny-path mock test with `GITLAB_ALLOWED_PROJECT_IDS`.

index.ts
```diff
@@ -9020,6 +9024,264 @@ async function purgeDependencyProxyCache(groupId: string): Promise<void> {
...
+
+  if (data.vulnerabilityDismiss.errors?.length) {
+    throw new Error(
+      `Failed to dismiss vulnerability: ${data.vulnerabilityDismiss.errors.join(", ")}`
+    );
+  }
+  if (!data.vulnerabilityDismiss.vulnerability) {
+    throw new Error(`Vulnerability not returned after dismissal (id: ${vulnerabilityId})`);
+  }
+  return data.vulnerabilityDismiss.vulnerability;
```
zereight, line 9314: Same allowlist gap as get — and this one writes. Don't call `vulnerabilityDismiss` until the vuln's project is checked against `GITLAB_ALLOWED_PROJECT_IDS`.

index.ts
```diff
@@ -9020,6 +9024,264 @@ async function purgeDependencyProxyCache(groupId: string): Promise<void> {
...
+
+  if (data.vulnerabilityConfirm.errors?.length) {
+    throw new Error(
+      `Failed to confirm vulnerability: ${data.vulnerabilityConfirm.errors.join(", ")}`
+    );
+  }
+  if (!data.vulnerabilityConfirm.vulnerability) {
+    throw new Error(`Vulnerability not returned after confirmation (id: ${vulnerabilityId})`);
+  }
+  return data.vulnerabilityConfirm.vulnerability;
```
zereight, line 9352: Same thing for confirm — check project ownership / allowlist before `vulnerabilityConfirm`.

zereight (changes requested):
Looks good overall — registry/skill sync and readonly gating are fine.

One blocker: `get_vulnerability` / `dismiss_vulnerability` / `confirm_vulnerability` only take a vulnerability id and skip `GITLAB_ALLOWED_PROJECT_IDS`. List is gated, these three aren't. Same hole we closed for `get_draft_note` in #598.

Also README / `discover_tools` still doesn't list the `vulnerabilities` category (docs/skill already do).

Minor: `toVulnerabilityGid` will pass through non-Vulnerability GIDs — worth a quick allowlist check and a deny-path test.

Please fix the allowlist gap before merge.

sidi7:
Thanks a lot for the careful review!

All three points fixed in 3f48d8e:

- Allowlist: get/dismiss/confirm now check the vuln's actual project against GITLAB_ALLOWED_PROJECT_IDS. I went with verify-after-fetch instead of a project_id param.
- README: now lists the four tools and the vulnerabilities category
- toVulnerabilityGid: only accepts numeric IDs or gid://gitlab/Vulnerability/<id> now

zereight (approved):
Reviewed 3f48d8e: allowlist enforcement on get/dismiss/confirm, GID validation, README/discover_tools update. LGTM.

Issue #575, opened 2026-07-04
Response schema rejects valid GitLab responses — `*_access_levels[].user_id/.group_id` and `commit` are non-nullable but GitLab returns `null`


prosomoalexd:
## Summary

Several tools fail with Zod validation errors on **well-formed** GitLab API responses. The response schemas mark fields as non-nullable that GitLab legitimately returns as `null`, so valid responses are rejected before reaching the caller. For `protect_branch` this is worse: the mutation is applied server-side, then the *response* fails to parse — so a successful write surfaces as an error.

## Affected tools

- `list_protected_branches`
- `protect_branch`
- `get_branch_diffs`

(and any tool sharing the protected-branch access-level or branch-diff commit schemas)

## Environment

- GitLab.com REST API
- Response validation via Zod
- Triggered with **role-based** protected-branch access levels (e.g. "Maintainers"), not user-/group-specific ones

## Steps to reproduce

1. Protect a branch using a **role-based** push/merge access level (e.g. Maintainers) rather than a specific user or group.
2. Call `list_protected_branches` (or `protect_branch` on that branch).
3. Separately, call `get_branch_diffs` over a range where GitLab returns a `null` commit field (e.g. an empty/no-op range).

## Actual behavior

`list_protected_branches`:

```
Invalid arguments: 0.push_access_levels.0.user_id: Expected number, received null,
0.push_access_levels.0.group_id: Expected number, received null ...
```

`protect_branch(develop)` — same error, **but the protection was actually applied** (a subsequent `list_branches` shows `protected: true`):

```
Invalid arguments: 0.push_access_levels.0.user_id: Expected number, received null ...
```

`get_branch_diffs(from=<feature-branch>, to=develop)`:

```
Invalid arguments: commit: Expected object, received null
```

Root cause: GitLab returns `null` for `user_id`/`group_id` on **role-based** access levels (only user-/group-specific levels populate them), and can return a `null` `commit` on branch-diff responses — but the schema marks these non-nullable.

## Impact

- Blocks **reading** protected-branch state and branch diffs entirely.
- Makes protected-branch **writes** look failed *after they have already applied* — inviting duplicate/retry attempts.
- Not workflow-specific: affects every consumer of these tools.

## Expected behavior

Valid GitLab responses parse successfully. In particular:

- `*_access_levels[].user_id` and `*_access_levels[].group_id` accept `null`.
- The branch-diff `commit` field accepts `null`.
[...]

PR #639, opened 2026-08-05
fix: accept null merged_yaml/includes in CI lint result

### LINKED ISSUE #638: validate_ci_lint throws on every invalid config: merged_yaml/includes are null, schema uses .optional()
## Summary

`validate_ci_lint` throws instead of returning a result **whenever the CI config is invalid** — that is, in exactly the case the tool is useful for. On a valid config it works fine, so the bug stays invisible until the user actually has a broken `.gitlab-ci.yml`.

## Cause

GitLab returns `merged_yaml: null` and `includes: null` when the config is invalid. `GitLabCiLintResultSchema` declares both as `.optional()`:

https://github.com/zereight/gitlab-mcp/blob/main/schemas.ts#L317-L324

```ts
export const GitLabCiLintResultSchema = z.object({
  valid: z.coerce.boolean(),
  errors: z.array(z.string()),
  warnings: z.array(z.string()).optional(),
  merged_yaml: z.string().optional(),
  includes: z.array(z.unknown()).optional(),
  jobs: z.array(z.unknown()).optional(),
});
```

Zod's `.optional()` accepts `undefined` but **rejects `null`**, so `.parse()` throws in `validateCiLint`/`validateProjectCiLint` and the tool surfaces a schema error rather than the lint result.

This is the same class of bug as #575 (role-based protected-branch access levels), fixed there by allowing `null`.

## Reproduction

Any invalid config will do. This one uses a job named `image`, which collides with the `default: image:` block:

```yaml
default:
  image: alpine:3.20

stages:
  - build

image:
  stage: build
  script:
    - echo build
```

**Direct API** — `POST /projects/:id/ci/lint`:

```json
{"valid":false,"errors":["image is defined in top-level and `default:` entry"],"warnings":[],"merged_yaml":null,"includes":null}
```

**Through the MCP tool** — `validate_ci_lint` with the same content:

```json
{"code":-32603,"message":"Invalid arguments: merged_yaml: Expected string, received null, includes: Expected array, received null"}
```

The actual explanation — ``image is defined in top-level and `default:` entry`` — never reaches the caller.

## Expected

The tool returns the lint result with `valid: false` and the `errors` array, so the caller can see why the config is rejected.

## Impact

A coding agent using this tool to validate CI changes cannot tell "the config is invalid" from "the tool is broken". In our case the agent concluded the tool was broken, silently fell back to `prettier --check` — which validates YAML formatting, not GitLab CI semantics — and reported success on a config that was invalid. Both error classes it needed to catch (a job named `image`, and `needs` pointing at a nonexistent job) are caught by the lint endpoint in under a second and are invisible to a formatter.

## Environment

- `@zereight/mcp-gitlab` 2.1.42 and 2.1.46 — reproduced on both
- self-hosted GitLab 19.2.0-ee
- transport: stdio, PAT auth, `GITLAB_PERMISSION_MODE=modify`
- client: Claude Code

## Suggested fix

`.nullish()` on the two fields GitLab actually nulls:

```diff
-  merged_yaml: z.string().optional(),
-  includes: z.array(z.unknown()).optional(),
+  merged_yaml: z.string().nullish(),
+  includes: z.array(z.unknown()).nullish(),
```

Happy to open a PR with a regression test in `test/nullable-gitlab-response-fields.test.ts` alongside the #575 cases.

nkz-soft:
Fixes #638.

## Problem

`validate_ci_lint` (and `validate_project_ci_lint`) throw instead of returning a result **whenever the CI config is invalid** — exactly the case the tool is called for. On a valid config they work, which is why the bug stays invisible until a user actually has a broken `.gitlab-ci.yml`.

GitLab returns `merged_yaml: null` and `includes: null` for an invalid config. `GitLabCiLintResultSchema` declared both as `.optional()`, which in Zod accepts `undefined` but rejects `null`, so `.parse()` threw and the caller got a schema error instead of the lint result:

```
-32603  Invalid arguments: merged_yaml: Expected string, received null, includes: Expected array, received null
```

The actual explanation from GitLab — e.g. ``image is defined in top-level and `default:` entry`` — never reached the caller.

## Change

`.optional()` → `.nullish()` on the two fields GitLab actually nulls. Same class as #575, so the regression test goes next to those cases in `test/nullable-gitlab-response-fields.test.ts`.

Scope is deliberately minimal: `warnings` and `jobs` come back as arrays in every response I observed, so I left them alone rather than widening the schema speculatively.

## Verification

The new test reproduces the reported failure before the schema change:

```
not ok 3 - ci lint result accepts null merged_yaml/includes on an invalid config
        "message": "Expected string, received null"
# pass 2
# fail 1
```

and passes after it. All three CI gates pass locally:

| Gate | Result |
|---|---|
| `npm run test:mock` | exit 0, 66 suites, 0 failures |
| `npm run test:consumer-smoke` | `Consumer install smoke passed.` |
| `npx tsc --noEmit` | clean |

Reproduced against a self-hosted GitLab 19.2.0-ee on 2.1.42 and 2.1.46, stdio transport with PAT auth, using this config:

```yaml
default:
  image: alpine:3.20

stages:
  - build

image:
  stage: build
  script:
    - echo build
```

Direct API response for it:

```json
{"valid":false,"errors":["image is defined in top-level and `default:` entry"],"warnings":[],"merged_yaml":null,"includes":null}
```

## Why it matters

An agent using this tool to validate CI changes cannot distinguish "the config is invalid" from "the tool is broken". In our case the agent concluded the latter, fell back to `prettier --check` — which checks YAML formatting, not GitLab CI semantics — and reported success on a config that was invalid. The two errors it needed to catch (a job named `image`, and `needs` referencing a nonexistent job) are both caught by the lint endpoint in under a second and are both invisible to a formatter.

### DIFF
--- a/schemas.ts
+++ b/schemas.ts
@@ -318,8 +318,10 @@ export const GitLabCiLintResultSchema = z.object({
   valid: z.coerce.boolean(),
   errors: z.array(z.string()),
   warnings: z.array(z.string()).optional(),
-  merged_yaml: z.string().optional(),
-  includes: z.array(z.unknown()).optional(),
+  // GitLab sends null (not an omitted field) for both when the config is
+  // invalid, which is exactly when this tool is called. See #638.
+  merged_yaml: z.string().nullish(),
+  includes: z.array(z.unknown()).nullish(),
   jobs: z.array(z.unknown()).optional(),
 });
 
--- a/test/nullable-gitlab-response-fields.test.ts
+++ b/test/nullable-gitlab-response-fields.test.ts
@@ -1,6 +1,7 @@
 import assert from "node:assert/strict";
 import { test } from "node:test";
 import {
+  GitLabCiLintResultSchema,
   GitLabCompareResultSchema,
   GitLabProtectedBranchAccessLevelSchema,
 } from "../schemas.js";
@@ -24,4 +25,20 @@ test("branch compare accepts null commit", () => {
     diffs: [],
   });
   assert.equal(parsed.commit, null);
+});
+
+test("ci lint result accepts null merged_yaml/includes on an invalid config", () => {
+  // #638: GitLab nulls both fields whenever the config is invalid — i.e. in
+  // exactly the case validate_ci_lint exists to report on.
+  const parsed = GitLabCiLintResultSchema.parse({
+    valid: false,
+    errors: ["image is defined in top-level and `default:` entry"],
+    warnings: [],
+    merged_yaml: null,
+    includes: null,
+  });
+  assert.equal(parsed.valid, false);
+  assert.equal(parsed.merged_yaml, null);
+  assert.equal(parsed.includes, null);
+  assert.deepEqual(parsed.errors, ["image is defined in top-level and `default:` entry"]);
 });


zereight (approved):
LGTM. Tested against null responses for merged_yaml and includes on invalid configs.

PR #644, opened 2026-08-08
fix: path traversal in download_job_artifacts local_path

brennanneoh:
## Summary
- `downloadJobArtifacts()` (`index.ts`, backing the `download_job_artifacts` tool) joined the caller-supplied `local_path` into the filesystem save path with no validation, allowing an absolute path or `../` traversal to write the downloaded artifact zip anywhere the process has write permission.
- The sibling `downloadAttachment()` function already validates its own `local_path` the same way; this brings `downloadJobArtifacts` in line with that existing, hardened pattern (reject absolute paths and `..` traversal).

## Why
Security issue: an MCP client (or anything able to issue tool calls in local/stdio mode) could pass `local_path: "/etc/cron.d"` or `local_path: "../../../../home/user/.ssh"` to `download_job_artifacts` and get an arbitrary file write on the host.

Reported privately to the maintainer alongside a second, unrelated finding before opening this PR.

## How tested
- `npx tsc --noEmit`
- `npm run test:mock` (all suites pass; added two new tests to `test/test-job-artifacts.ts` covering traversal and absolute-path rejection, and updated two existing tests that relied on an absolute `local_path` to use a relative one, matching the contract already established by `downloadAttachment`'s test suite)
- `npm run test:consumer-smoke`

## Breaking changes
`local_path` for `download_job_artifacts` must now be a relative, non-traversal path — matching the existing contract for `download_attachment`'s `local_path`. Absolute paths or `..` segments now return an error instead of silently writing outside the intended location.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -7486,7 +7486,21 @@ async function downloadJobArtifacts(
   await handleGitLabError(response);
 
   const filename = `artifacts_job_${encodeGitLabPathSegment(jobId)}.zip`;
-  const savePath = localPath ? path.join(localPath, filename) : filename;
+  let savePath: string;
+  if (localPath) {
+    const normalizedLocalPath = path.normalize(localPath);
+    if (
+      path.isAbsolute(normalizedLocalPath) ||
+      normalizedLocalPath === ".." ||
+      normalizedLocalPath.startsWith(".." + path.sep) ||
+      normalizedLocalPath.includes(path.sep + ".." + path.sep)
+    ) {
+      throw new Error("Invalid local_path: directory traversal is not allowed.");
+    }
+    savePath = path.join(normalizedLocalPath, filename);
+  } else {
+    savePath = filename;
+  }
   fs.mkdirSync(path.dirname(savePath), { recursive: true });
 
   if (!response.body) {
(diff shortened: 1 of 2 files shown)


brennanneoh:
Pushed 4426f29 addressing the 3 actionable comments:
- moved `local_path` validation before the GitLab fetch
- added a check for Windows drive-relative paths (`path.parse().root`)
- updated `test/test-job-artifacts.ts` to use relative fixture dirs, plus added traversal + absolute-path rejection tests

On the symlink note: agreed it's a real gap, but it's pre-existing in `downloadAttachment` too (same lexical-only check), so I kept this PR scoped to bringing `downloadJobArtifacts` in line with that existing pattern rather than introducing new canonicalization/no-follow behavior across both. Happy to file that as a separate follow-up if useful.

brennanneoh:
Pushed bbd734f: fixed the flagged issue in the two new traversal-rejection tests. \`assert.fail()\` was inside the same \`try\` its own \`catch\` handled, so an unexpected success would have been silently swallowed as a pass. Now the error is captured outside the \`try/catch\` and asserted on there, so a non-throwing call correctly fails the test.

zereight:
Thanks for the PR! This issue and the broader path traversal / symlink escape vectors were addressed across all file download endpoints in #642 using `openSafeOutputWriteStream`. Since main already contains comprehensive validation and tests for `download_job_artifacts`, closing this PR as superseded.

PR #700, opened 2026-08-31
fix: honor configured npm registry in version-check

### LINKED ISSUE #699: Version-check hardcodes registry.npmjs.org, ignoring configured npm registry
`utils/version-check.ts` hardcodes the update-check URL:

```ts
const NPM_REGISTRY_LATEST_URL = "https://registry.npmjs.org/@zereight/mcp-gitlab/latest";
```

`fetchLatestVersion()` always fetches this literal URL directly, with no way to point it at a configured/private npm registry (no `npm_config_registry` / `NPM_CONFIG_REGISTRY` env var lookup, no `npm config get registry` call, no CLI flag).

**Impact:** in corporate environments where direct access to the public npm registry is blocked by network policy (proxy/EDR/DLP — e.g. Microsoft Defender for Endpoint's Network Protection), every server startup makes an outbound connection attempt to `registry.npmjs.org` that gets blocked. The check fails silently by design, so functionality isn't affected, but it generates a security-tooling alert/block notification on every single startup, which is confusing and noisy for users on a locked-down/private-registry setup (e.g. an internal Artifactory npm mirror configured via `.npmrc`).

`GITLAB_DISABLE_VERSION_CHECK=true` / `--disable-version-check=true` works around this by skipping the check entirely, but that also loses the "new version available" notice.

**Suggested fix:** resolve the registry the same way `.npmrc`-aware tooling does before falling back to the public default — e.g. `chrome-devtools-mcp` handles this by first trying `npm config get registry` and only falling back to `https://registry.npmjs.org` if that fails:

```js
function getRegistry() {
  try {
    const registry = execSync('npm config get registry', {
      encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'], timeout: 5000, windowsHide: true,
    }).trim().replace(/\/$/, '');
    if (registry && registry !== 'undefined' && /^https?:\/\//.test(registry)) return registry;
  } catch {}
  return 'https://registry.npmjs.org';
}
```

This would let the update check honor `.npmrc`/`npm_config_registry` in restricted environments while keeping the same fallback behavior everywhere else.

**Repro:**
1. Configure a corporate npm registry mirror in `~/.npmrc` (`registry=https://your-internal-mirror/...`).
2. Have network policy block direct outbound access to `registry.npmjs.org`.
3. Run `npx -y @zereight/mcp-gitlab` (e.g. as a Claude Code MCP server).
4. Observe a blocked outbound connection attempt to `registry.npmjs.org` on every startup, even though the configured registry mirror is never consulted.

zereight:
## Summary
- `fetchLatestVersion()` hardcoded `https://registry.npmjs.org`, ignoring any configured private/internal npm registry (`.npmrc`, `npm_config_registry`).
- In locked-down environments this made every server startup attempt a direct outbound call to the public registry, which network policy (proxy/EDR/DLP) blocks and alerts on — noisy even though the check fails silently by design and doesn't affect functionality.
- Now resolves the registry via `npm config get registry` first, falling back to the public registry when that lookup fails or returns something unusable.

Fixes #699

## Test plan
- [x] `npm run build`
- [x] `node --test build/test/utils/version-check.test.js` — added cases covering resolver override and fallback-on-throw

### DIFF
--- a/test/utils/version-check.test.ts
+++ b/test/utils/version-check.test.ts
@@ -33,12 +33,16 @@ function throwingResolver(): string {
   throw new Error("npm not found");
 }
 
-async function withNpmrc(contents: string, run: () => Promise<void>): Promise<void> {
+async function withNpmrc(
+  contents: string,
+  run: () => Promise<void>,
+  globalContents = ""
+): Promise<void> {
   const configDir = mkdtempSync(join(tmpdir(), "gitlab-mcp-npmrc-"));
   const userconfigPath = join(configDir, ".npmrc");
   const globalconfigPath = join(configDir, "global.npmrc");
   writeFileSync(userconfigPath, contents);
-  writeFileSync(globalconfigPath, "");
+  writeFileSync(globalconfigPath, globalContents);
 
   const previousUserconfig = process.env.NPM_CONFIG_USERCONFIG;
   const previousGlobalconfig = process.env.NPM_CONFIG_GLOBALCONFIG;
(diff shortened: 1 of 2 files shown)


zereight:
@coderabbitai review

zereight:
@coderabbitai full review

zereight:
@coderabbitai review

zereight:
@coderabbitai review

zereight:
@coderabbitai review

PR #710, opened 2026-08-31
feat: add ENABLE_DYNAMIC_PROJECT_SCOPE and ENABLE_STRICT_PROJECT_SCOPE

sovdchains:
Adds two opt-in flags for multi-tenant deployments.

`ENABLE_DYNAMIC_PROJECT_SCOPE` (needs `REMOTE_AUTHORIZATION=true`): a client can send an `X-GitLab-Allowed-Project-Ids` header to restrict its own session to a subset of projects. My use case is a chat bot talking to one gitlab-mcp instance where each room maps to a different project; without this I'd have to run a container per room. The header can only restrict further, never widen: if `GITLAB_ALLOWED_PROJECT_IDS` is set and the header names anything outside it, the request gets a 401.

`ENABLE_STRICT_PROJECT_SCOPE`: closes the discovery gaps an allowlist leaves open. Today the allowlist only guards tools that take a project_id, listings and searches still return everything the token can see. With this flag, project listings and `search_repositories` are filtered to the allowlist, `list_issues`/`list_merge_requests` without a project_id resolve to the allowed project instead of the instance-wide endpoint, and tools that can't be limited to allowed projects (instance-wide code search, todos, the group tools) are refused.

Both flags off: nothing changes. Tests and docs included, test suite passes.

One thing to note is that I reworded two error messages from "when GITLAB_ALLOWED_PROJECT_IDS is set" to "while a project allowlist is in effect (GITLAB_ALLOWED_PROJECT_IDS or a session scope)" because the allowlist can now come from the session.

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -123,6 +123,8 @@ export const ENABLE_DYNAMIC_API_URL =
   getConfig("enable-dynamic-api-url", "ENABLE_DYNAMIC_API_URL") === "true";
 export const ENABLE_DYNAMIC_PROJECT_SCOPE =
   getConfig("enable-dynamic-project-scope", "ENABLE_DYNAMIC_PROJECT_SCOPE") === "true";
+export const ENABLE_STRICT_PROJECT_SCOPE =
+  getConfig("enable-strict-project-scope", "ENABLE_STRICT_PROJECT_SCOPE") === "true";
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


sovdchains:
@coderabbitai review

sovdchains:
@coderabbitai i force pushed the changes you asked for. why still changes requested?

zereight (approved):
Approved. Both flags are off by default, widening fails closed, and the trust model is documented.

PR #721, opened 2026-09-03
Package gitlab-mcp for Nix

m4r1vs:
Hello, a coworker recommended your MCP. Project seems great. I use NixOS and home-manager to configure claude code. This PR adds a flake.nix file that defines how gitlab-mcp is built. This means, it can be imported in a flake directly and kept up-to-date to this upstream. Happy to answer any questions that might arise

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+# Loads the Nix devShell from flake.nix (requires direnv + nix-direnv).
+# Run `direnv allow` once after cloning.
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,7 @@ docs/plans/
 site/
 .venv-docs/
 .direnv/
+
+# nix build output
+result
+result-*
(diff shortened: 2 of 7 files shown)


zereight:
@coderabbitai review

zereight (dismissed):
Nice addition — NixOS/home-manager users get a clean install path now. Two small nits, neither blocking:

1. `systems` does not include `x86_64-darwin`. Intentional? If so a one-line comment would help; `nodejs_22` supports it, so worth adding otherwise.

2. `mainProgram = "mcp-gitlab"` while the READMEs point everyone at the `zereight-mcp-gitlab` alias — `lib.getExe` resolves to the legacy name. Aligning the two would avoid confusion.

m4r1vs:
Hey, thanks for your quick chime-in! I excluded x86 darwin because support for it will be dropped in the next nix release 26.11 (and has already been dropped on unstable). Using the current release with x86 darwin seems perfectly fine to me too tho

m4r1vs:
Good find regarding the binary name, I'll fix

PR #732, opened 2026-09-06
fix: defer OAuth token acquisition to first use; no browser on transient refresh errors

zereight:
## Problem

When `GITLAB_USE_OAUTH=true`, the server acquires the OAuth token **eagerly and synchronously at startup** (`runServer` → `initializeOAuthClient` → `getAccessToken`), blocking the MCP handshake on a network call to GitLab.

For a self-hosted GitLab (often behind a VPN), the network frequently isn't ready at the instant the MCP client spawns the server. The refresh `fetch` then hangs past the client's connection timeout, and on *any* refresh error the code falls through to the interactive `startOAuthFlow()` — **opening a browser on every cold start**.

## Fix

- **Defer token acquisition.** Construct the OAuth client synchronously at startup (no network) via a new `createGitLabOAuthClient`. The token resolves lazily on the first tool call (every call already passes through `ensureValidOAuthToken`), by which point the network is ready. `initializeOAuthClient` is kept and now delegates.
- **Always deliver the token (fixes #495 review feedback).** New testable helper `ensureOAuthToken(client, currentToken, onToken)`: the fast path requires a held token AND a valid stored token, so the first call after lazy startup populates the token even when the on-disk token is still valid. Token-script setups resolve exactly once.
- **Fast-fail refresh.** 10s `AbortSignal.timeout` on the refresh request.
- **Browser only when warranted.** `isAuthInvalidTokenResponse(status, body)` returns true only for 400/401 with `invalid_grant`/`invalid_token`. Transient and config errors propagate to the caller without a popup.

## Tests

- New `test/oauth-startup.test.ts` (14 cases): 8 classifier cases + 6 `ensureOAuthToken` cases (valid-cache delivery without network via closed-port URL, no re-delivery fast path, refresh + persist against a loopback `/oauth/token` mock, transient-failure propagation, token-script single resolution). Wired into the pure-unit group in `scripts/run-mock-tests.sh`.
- `npm run build` clean; full `npm run test:mock` green; `test-auth-retry` 19/19 and `oauth-device-flow` 22/22 still green.

## Behavior change

First tool call (instead of startup) opens the browser when no stored token exists; a transient refresh failure fails that call instead of popping a browser. Documented in `docs/configuration/environment-variables.md`.

---

Supersedes #495. Thanks to @medyas for the original fix — this PR carries it forward with the cached-token fix and regression tests from the review.

### DIFF
--- a/docs/configuration/environment-variables.md
+++ b/docs/configuration/environment-variables.md
@@ -25,6 +25,10 @@ Optional Netscape cookie file path for cookie-based authentication.
 
 Set to `true` to enable local browser-based OAuth2 authentication.
 
+The token is acquired lazily on the first tool call (not at startup). If no
+stored token exists, the browser opens then; transient refresh failures fail
+that call instead of opening a browser.
+
 ### `GITLAB_OAUTH_CLIENT_ID`
 
 Client ID for local OAuth2 authentication.
--- a/index.ts
+++ b/index.ts
@@ -144,7 +144,7 @@ import { CookieJar, parse as parseCookie } from "tough-cookie";
 import { URL } from "node:url";
 import { z } from "zod";
 
-import { initializeOAuthClient, GitLabOAuth } from "./oauth.js";
+import { createGitLabOAuthClient, ensureOAuthToken, GitLabOAuth } from "./oauth.js";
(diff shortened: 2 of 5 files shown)


zereight:
@cursor please review this PR

PR #741, opened 2026-09-13
fix: omit empty merge request list filters

### LINKED ISSUE #740: list_merge_requests sends empty optional filters and receives 400
## Problem

Some MCP clients send unfilled optional `list_merge_requests` arguments as empty strings or empty arrays. The server serializes those values into the GitLab REST query instead of omitting them.

For example, an otherwise unfiltered call can produce:

```text
/api/v4/merge_requests?assignee_id=&assignee_username=&author_id=&author_username=&reviewer_id=&reviewer_username=&labels=&milestone=&search=&target_branch=&source_branch=
```

GitLab treats the simultaneously present ID and username filters as mutually exclusive and responds with `400 Bad Request`.

## Root cause

`cleanMutuallyExclusiveIdUsernameOptions()` only resolves conflicts when a username filter has a non-empty value. Both empty values therefore remain. `appendMergeRequestFilters()` currently skips only `undefined`, so it appends empty strings and turns an empty `labels` array into `labels=`.

## Expected behavior

- Empty strings and empty arrays are omitted when serializing merge request list filters.
- Non-empty filters keep their current encoding.
- Meaningful falsy values such as `false` and `0` remain supported.
- Both project and group merge request list endpoints remain compatible because they share the serializer.

## Reproduction test

Add an MCP-level test that calls `list_merge_requests` with empty optional filters and uses a mock GitLab endpoint that returns 400 if any empty query value is received. The final request should be exactly `/api/v4/merge_requests`.

Danny5487401:
## Summary

- omit empty and whitespace-only strings, plus empty arrays, when serializing merge request list filters
- prevent empty mutually exclusive ID and username parameters from reaching GitLab and causing HTTP 400
- add an MCP-level regression test covering spaces, tabs, newlines, empty strings, and empty arrays
- keep non-empty filters and meaningful falsy values unchanged

Fixes #740

## Tests

- `npm run test:list-merge-requests` — 5 passed
- `npm run build` — passed
- `npx tsx test/test-list-group-merge-requests.ts` — 3 passed
- `node --import tsx/esm --test test/utils/tool-args.test.ts` — 9 passed

## Existing repository test limitations

- The full Mock suite requires Node.js 22 because Node.js 21 does not support `--experimental-test-isolation=none`.
- Under Node.js 22, the suite reaches an existing unrelated failure in `remote-auth-simple-test.ts`: expected expired Session status 404, received 406.
- `npm run lint` cannot start on current main because ESLint 9 cannot find an `eslint.config.*` file.


docs/ai-changes-gitlab-mcp-list-merge-requests-empty-filter-fix.md
```diff
@@ -0,0 +1,39 @@
+# 修复 list_merge_requests 空筛选参数
```
zereight, line 1: Can we drop this file from the PR? It reads like a working note from the session, and the PR description already covers the same ground. It's also not linked from the mkdocs nav, so it would sit orphaned in docs/.

index.ts
```diff
@@ -2409,7 +2409,11 @@ async function markAllTodosDone(): Promise<void> {
  */
 function appendMergeRequestFilters(url: URL, options: Record<string, unknown>): void {
   Object.entries(options).forEach(([key, value]) => {
-    if (value === undefined) {
+    if (
+      value === undefined ||
+      value === "" ||
```
zereight, line 2414: Should whitespace-only strings be skipped too? `" "` passes this guard and still gets serialized (e.g. `search=%20`). Maybe `typeof value === "string" && value.trim() === ""`.

zereight (changes requested):
Two small things: please drop the session note from docs/, and consider handling whitespace-only filter values. The core fix itself looks good.

docs/ai-changes-gitlab-mcp-list-merge-requests-empty-filter-fix.md
Danny5487401: Done in 0eb6db8. I removed the session working note from docs/ and updated the PR description so the fix remains documented there.

index.ts
Danny5487401: Done in 0eb6db8. The serializer now skips strings whose trim() result is empty, and the regression test covers spaces, tabs, and newlines.

Danny5487401:
Addressed both review comments in commit `0eb6db8`:

- Removed `docs/ai-changes-gitlab-mcp-list-merge-requests-empty-filter-fix.md` from the PR and kept the relevant context in the PR description.
- Updated merge-request filter serialization to omit whitespace-only strings using `value.trim() === ""`; the regression test now covers spaces, tabs, and newlines.

Validation passed:
- `npm run test:list-merge-requests` (5/5)
- `npx tsx test/test-list-group-merge-requests.ts` (3/3)
- `node --import tsx/esm --test test/utils/tool-args.test.ts` (9/9)
- `npm run build`

zereight (approved):
Both items addressed — docs note dropped, whitespace handling added. Thanks!

PR #751, opened 2026-09-15
feat: add configurable response masking for sensitive GitLab data

anime-fan-5903:
## What

Adds opt-in response masking for GitLab tool results and returned error messages.

When enabled, the server replaces sensitive values with mask text before they are returned to the MCP client. It provides built-in masking for GitLab tokens and IP addresses, plus configurable keyword and regular-expression rules for organization names, personal names, internal administrative endpoints, and other deployment-specific sensitive data.

Masking is applied at the MCP response boundary across stdio, SSE, and Streamable HTTP transports. Managed policy groups can also select server-owned masking rules by GitLab instance and project.

## Why

Internal GitLab projects commonly contain operational details in issues, merge requests, pipeline logs, and related records. This can include internal administration URLs, IP addresses, access tokens, company names, employee names, and other sensitive information. Their presence is often legitimate and does not indicate poor project hygiene.

When an MCP server is used with a cloud-hosted model, returning these values directly creates a material information-disclosure risk. Response masking lets teams preserve normal GitLab workflows while reducing the sensitive data exposed to an AI client.

## How tested

- Ran `npm run build`.
- Ran response-masking unit tests: 22 passing.
- Ran MCP boundary integration tests: 7 passing.
- Verified masking through Codex: when a tool response contains configured sensitive values, Codex receives only the masked replacement text.
- Verified behavior across stdio and Streamable HTTP, including managed policy selection for request-local GitLab instances.

## Current limitations and future considerations

Response masking currently applies only to text content, including MCP text blocks, JSON represented in text responses, structured content, and returned error messages. It does not inspect images, binary content, local files, download URLs, downloaded files, or base64-encoded payloads.

If this feature is accepted, future work should consider extending protection to additional response types where technically and operationally appropriate.

Masking rules are also not guaranteed to detect every sensitive value. Users may still face information-disclosure risk from content that does not match configured or built-in rules.

One possible complementary approach would be an intermediate review step using a lightweight local LLM before data is sent to a cloud-hosted model. That approach has meaningful cost, latency, reliability, and security trade-offs, so it should be evaluated separately rather than treated as a replacement for deterministic masking.

## Breaking changes

None.

Masking is disabled by default, so existing deployments retain their current response behavior unless `GITLAB_MASKING_ENABLED=true` is configured.


masking/engine.ts
```diff
@@ -0,0 +1,524 @@
...
+    options?: { textFormat?: "plain" | "json" }
+  ): T;
+  maskError(error: unknown): unknown;
+}
+
+const BUILTIN_RULES: MaskRule[] = [
+  {
+    id: "gitlab-token",
+    type: "regex",
+    pattern: "\\bgl(?:pat|rt|pt|ft|imt|agent)-[A-Za-z0-9_-]{20,}\\b",
```
zereight: Builtin token pattern covers 6 prefixes (`pat|rt|pt|ft|imt|agent`) but misses documented families like `gldt-` (deploy), `glptt-` (pipeline trigger), `glcbt-` (CI job), `gloas-`/`glsoat-` (OAuth/SCIM) — those would pass through silently. Note `glpt-` here also doesn't match the real trigger prefix `glptt-`. Suggest widening the alternation and adding one test per covered prefix in `test/utils/masking.test.ts` (currently only `glpat-` is covered).

docs/configuration/environment-variables.md
```diff
@@ -7,6 +7,86 @@ Use an API URL, not the GitLab web root:
...
+
+### `GITLAB_MASKING_WORKSPACE_DIR`
+
+Optional directory used to resolve the masking configuration. This does not
+change the process working directory. For local stdio use, set it to the
+workspace directory when the MCP client does not set the server process cwd.
+
+The config file supports `keyword` rules for literal replacements and `regex`
+rules for pattern replacements. See `.gitlab-mcp-mask.example.json` in the
+repository. When masking is enabled, built-in rules cover GitLab token formats,
```
zereight: "built-in rules cover GitLab token formats" reads as complete coverage, but the builtin only matches 6 `gl*` prefixes (and misses e.g. `gldt-`/`glptt-`; see `masking/engine.ts`). Suggest enumerating the covered prefixes here so operators know what's actually masked.

index.ts
```diff
@@ -853,31 +871,88 @@ function createServer(): McpServer {
...
+          }
+        : undefined;
+      const selectMaskingPolicy = () => {
+        const isToolDiscovery = toolName === "discover_tools";
+        const inputSchema = allTools.find(tool => tool.name === toolName)?.inputSchema;
+        // Optional project fields can be filters on global queries. Only supply
+        // the default when the corresponding handler actually uses it.
+        const usesDefaultProject =
+          inputSchema?.required?.includes("project_id") ||
+          toolName === "get_file_contents" ||
```
zereight, line 932: This hardcodes which handlers fall back to the default project. If a handler's default-project behavior changes later, masking selection silently diverges (spurious deny or builtin fallback). Consider a shared helper or a comment linking both sides so they stay in sync.

test/response-masking.test.ts
```diff
@@ -0,0 +1,338 @@
...
+      LOG_LEVEL: "error",
+      ...options.env,
+    },
+  });
+  transport.stderr?.on("data", () => undefined);
+  await client.connect(transport);
+  return { client, requests };
+}
+
+describe("managed response masking at the MCP boundary", { timeout: 30_000 }, () => {
```
zereight, line 160: All 7 cases run managed mode on the stateful path. Could you add one unmanaged-mode boundary case (local config file, no policy file) and one stateless-mode case? The docs claim stateless is covered, and that path currently has no execution evidence.

zereight (changes requested):
Thanks for this — opt-in default, fail-closed selection, and boundary placement across transports all look right, and the managed-mode deny behavior is well tested (verified 22 unit + 7 boundary green locally on this head).

Requesting changes for one gap: the builtin `gitlab-token` pattern only matches 6 prefixes while the docs claim full GitLab token coverage, so documented families like deploy/trigger/job tokens would pass through silently. Details inline, plus two smaller follow-ups (default-project list maintenance, unmanaged/stateless test coverage).

Note: CI workflows show `action_required` (fork approval gate) — needs maintainer approval to run.

anime-fan-5903:
Thank you for your response; I have created a fix commit.

masking/engine.ts
zereight: Fixed in 4cdf33a.

docs/configuration/environment-variables.md
zereight: Fixed in 4cdf33a.

index.ts
zereight: Fixed in 4cdf33a.

test/response-masking.test.ts
zereight: Fixed in 4cdf33a.

PR #756, opened 2026-09-19
fix: detect comma-separated mutations in the modify-mode delete guard

zereight:
## What

GraphQL treats commas as insignificant tokens, so these two documents are identical to GitLab:

```graphql
mutation M { issueDelete(input: { iid: "1" }) { id } }
, mutation M { issueDelete(input: { iid: "1" }) { id } }
```

The modify-mode guard `graphqlQueryContainsDeleteOperation()` only recognised a `mutation` keyword preceded by start-of-string, `;` or `}`, so a comma-prefixed (or comma-separated) mutation was classified as non-destructive and forwarded to the GitLab instance with the operator's token.

The read-only write check (`graphqlQueryContainsWriteOperation`) already handled commas; the delete check did not.

## Change

- Accept `,` as an operation separator in both the detection regex and `extractTopLevelMutationFields()` (`utils/graphql-query.ts`).
- Add tests for the comma-prefixed, comma-without-whitespace and comma-separated forms, plus a control that two comma-separated queries stay allowed.

## Verification

- `npm run build` passes.
- `node --import tsx/esm --test --experimental-test-isolation=none test/utils/graphql-query.test.ts` → 31/31 pass (4 new tests).
- `node --import tsx/esm --test --experimental-test-isolation=none test/test-permission-mode.ts` → 13/13 pass.

Reported through GitHub private vulnerability reporting. The same report also covers destructive mutations whose names do not contain a delete verb; that part is handled in a separate PR.


zereight:
@cursor review

zereight:
@cursor please review this PR

PR #758, opened 2026-09-19
fix: drop GitLab credentials on cross-origin download redirects

zereight:
Stacked on #757 (base branch `fix/download-redirect-destination-validation`); review that one first.

## What

The download sinks attach the server's GitLab credential to the outbound request:

- `registerDownloadProxy` sets `Private-Token` / `JOB-TOKEN` from the request or from the encrypted `_token` (`downloads/proxy.ts`), and in remote static-token mode that token is the server's own credential;
- `downloadReleaseAsset` uses the configured auth header for the tool call.

When the upstream answered with a redirect to another host, the HTTP client followed it and forwarded those custom headers. undici strips only `authorization`, `proxy-authorization`, `cookie` and `host` across origins, so `Private-Token` and `JOB-TOKEN` survived the hop:

```text
undici default (redirect: "follow"), cross-origin hop receives: glpat-…
with fetchWithValidatedRedirects():                          none
```

A GitLab release asset link whose `url` points at an attacker-controlled collector therefore delivered the server's raw GitLab credential to that host.

## Change

- In `fetchWithValidatedRedirects()` (`utils/safe-redirect-fetch.ts`), drop credential headers (`private-token`, `job-token`, `authorization`, `cookie`, case-insensitive) as soon as a redirect crosses to a different origin. Same-origin redirects keep them.
- Expose `stripCredentialHeaders()` and `DEFAULT_CREDENTIAL_HEADER_NAMES`; the header list is overridable per call via `credentialHeaderNames`.
- Tests: same-origin redirect still carries `Private-Token`; cross-origin redirect arrives without it.

## Verification

- `npm run build` passes.
- `node --import tsx/esm --test --experimental-test-isolation=none test/utils/safe-redirect-fetch.test.ts` → 9/9 pass (2 new tests).
- `test/utils/*.test.ts` → 145/145 pass.

Reported through GitHub private vulnerability reporting.


zereight:
@cursor review

zereight:
@cursor please review this PR

zereight:
Closing this as superseded by #757.

The credential strip it introduces is already in #757 in a more complete form: `CREDENTIAL_HEADERS` / `withoutCredentialHeaders` on every hop that leaves the originating origin, plus `credentialsWithheld` so a stripped hop is neither re-authenticated by a later hop back to the origin nor routed through the cookie-jar / OAuth-retry wrapped client, and `unauthenticatedFetchImpl` defaulting to plain `undici`. The tests for it live in `test/utils/safe-redirect-fetch.test.ts` (the header policy across hops, trusted-host opt-in, the per-hop client split).

This PR was stacked on `fix/download-redirect-destination-validation` and only carried those two files, which is why it went `CONFLICTING` as #757 moved. Cursor's review of #757 confirms the reported vulnerability — custom headers surviving a cross-origin redirect, since `undici` strips only `authorization`, `proxy-authorization`, `cookie` and `host` — is closed there. Nothing from this branch is lost.

PR #759, opened 2026-09-19
fix: apply session limits to the SSE transport

zereight:
## What

`GET /sse` (`startSSEServer`) allocated a `Server` instance, an `SSEServerTransport` and a held response socket for every connection, with no capacity check, no per-IP rate limiter and no idle timeout. A client that can reach the endpoint can therefore open never-closing streams until the single-process Node server runs out of memory or file descriptors, denying service to every other client — including in the "recommended secure" configuration where `SSE_AUTH_TOKEN` is set, since the attacker only needs the shared token.

The Streamable HTTP transport already has all three controls (`MAX_SESSIONS` capacity gate, per-IP rate limit, idle session expiry); the SSE transport never got them.

## Change

In `startSSEServer()` (`index.ts`):

- **capacity**: new `GET /sse` connections are rejected with `503` (`Server capacity reached`) once `MAX_SESSIONS` sessions are active, mirroring `/mcp`;
- **rate limit**: connection creation is limited per client IP via `express-rate-limit` with the same key generator `/mcp` uses (`MAX_REQUESTS_PER_MINUTE`, default 60); `POST /messages` is intentionally not limited so existing SSE clients keep their message throughput;
- **idle timeout**: sessions that receive no `POST /messages` for `SESSION_TIMEOUT_SECONDS` are closed and their slot released, via an unref'd sweep interval that is cleared on shutdown;
- docs updated under `MAX_SESSIONS` and `SESSION_TIMEOUT_SECONDS`.

## Compatibility

SSE clients that hold a connection open and exchange messages keep working. A client that connects and sends nothing for `SESSION_TIMEOUT_SECONDS` (default 1 hour) is disconnected and can reconnect.

## Verification

- `npm run build` passes.
- New `test/sse-session-limits.test.ts` → 2/2 pass (third connection over `MAX_SESSIONS=2` gets 503; an idle connection with `SESSION_TIMEOUT_SECONDS=2` is closed by the server).
- `test/sse-auth-guard.test.ts` → 2/2, `test/streamable-http-sse-stream.test.ts` → 8/8.

Reported through GitHub private vulnerability reporting.


zereight:
@cursor review

zereight:
@cursor please review this PR

zereight:
@cursor please review this PR

PR #753, opened 2026-09-19
fix: drop blank entries inside merge request list array filters

zereight:
Follow-up to #741, extended to the remaining list tools.

Blank filters were leaking into request URLs. #741 handled empty scalars and empty arrays, the next commit handled blank entries *inside* merge request arrays, and the one after that applied the guard to the issue and milestone list endpoints through shared helpers. This branch takes the same guard to every other query builder in `index.ts`, so the scope now matches the title.

## Change

- `utils/tool-args.ts`: new `dropBlankArrayEntries` / `isBlankFilterValue` helpers, plus `appendFilterParam` / `appendFilterParams` for query serialization.
- `index.ts`:
  - `appendMergeRequestFilters` now uses the shared helpers.
  - `list_issues`: `labels: ["", "bug"]` keeps only `bug`; all-blank arrays and blank `assignee_username` entries are omitted.
  - `list_milestones` / `list_group_milestones`: `iids: []` is no longer serialized as `iids=`; blank scalars such as `title` and `search` are omitted.
  - 27 more query builders route through `appendFilterParams`, including the nested `filter[environment_scope]` values used by `list_project_variables` and `list_group_variables`.
  - `list_labels` no longer sends `project_id` as a query parameter on top of the path segment.
- `schemas.ts`: the `iids` description now says an empty array means "no filter".

## Before / after

| Tool | Input | Before | After |
|---|---|---|---|
| `list_merge_requests` | `labels: ["", "bug"]` | `labels=,bug` | `labels=bug` |
| `list_issues` | `labels: ["", "bug"]` | `labels[]=&labels[]=bug` | `labels[]=bug` |
| `list_milestones` | `iids: []` | `iids=` | omitted |
| `list_group_milestones` | `iids: []` | `iids=` | omitted |
| `list_labels` | `search: ""` | `search=` | omitted |
| `list_tags`, `list_pipelines`, `list_deployments`, `list_environments`, `list_job_artifacts`, `list_events` | blank string filter | `search=` etc. | omitted |
| `list_project_variables` | `filter: { environment_scope: "  " }` | `filter[environment_scope]=` | omitted |

## Tests

- `test/test-blank-filters.ts` (new): 8 tools end to end plus 2 control cases. Every blank case failed against the unfixed code.
- `test/test-list-milestones.ts` (new): 4 cases covering empty/blank milestone filters and the bracket-array `iids[]` form.
- `test/test-list-issues.ts`: 2 cases for blank array entries.
[...]

PR #763, opened 2026-09-20
fix: keep the id filter when the username filter is blank

zereight:
## Problem

`list_issues`, `list_merge_requests` and `list_group_merge_requests` resolve the
`*_id` / `*_username` conflict by dropping the id when a username is present, because
GitLab rejects both at once with a 400.

`hasUsernameFilterValue` decided "present" by truthiness:

```ts
if (Array.isArray(value)) {
  return value.length > 0;
}

return Boolean(value);
```

The query builder judges blanks differently: `appendFilterParam` drops blank array
entries and skips values that are blank afterwards. So `"  "` or `[""]` counted as a
username in the conflict guard, the id got deleted, and then the blank guard dropped
the username too. The request went out with no filter at all — a silently unfiltered
list instead of a 400.

## Fix

`hasUsernameFilterValue` now asks the same question as `appendFilterParam`:

```ts
if (value === undefined || value === null) {
  return false;
}

return !isBlankFilterValue(dropBlankArrayEntries(value));
```

The id filter is only deleted when a username actually survives to the query string —
which is exactly the combination GitLab answers with a 400.

## Behavior

Measured against `assignee_id: "7"` with the real predicate and the real query
builder, not asserted from reading:

| `assignee_username` | old result | new result | `appendFilterParam` output |
| --- | --- | --- | --- |
| `["bob"]` | id dropped | id dropped | `assignee_username=bob` |
| `["", "bob"]` | id dropped | id dropped | `assignee_username=bob` |
| `"  "` | id dropped, username dropped → **unfiltered** | id kept | *(dropped)* |
| `[""]` | id dropped, username dropped → **unfiltered** | id kept | *(dropped)* |
| `["", "  "]` | id dropped, username dropped → **unfiltered** | id kept | *(dropped)* |
| `""` | id kept | id kept | *(dropped)* |
| `[]` | id kept | id kept | *(dropped)* |
| `undefined` | id kept | id kept | *(dropped)* |
| `[0]`, `["", 0]` | id dropped | id dropped | `assignee_username=0` |

The three bold rows are the bug: the id was deleted while the blank guard removed the
username, so `list_issues` returned every issue instead of the filtered set. Nothing
else moves.

`author_*` and `reviewer_*` go through the same function, so they change together.

## Not in scope

`isBlankFilterValue(null)` is `false` — it checks `=== undefined` but not `null` — so
`appendFilterParam` would serialize a literal `assignee_username=null`. The `null` early
return in `hasUsernameFilterValue` keeps the id filter for that value, matching the old
behavior, but the serialization quirk stays. It is unreachable from the tool path:
`sanitizeToolArguments` strips top-level `null` before the schemas parse
(`index.ts:10842`). Separate PR if we want `isBlankFilterValue` to treat `null` as blank,
since that widens every tool's query building.

## Tests

Two cases added to `test/utils/tool-args.test.ts`:

- `assignee_username: ["", "  "]` keeps `assignee_id`
- `assignee_username: ["", "bob"]` still drops it

Existing cases for `[]`, omitted values and non-blank values are unchanged.

## Verification

```
npx tsc --noEmit
node --import tsx/esm --test --experimental-test-isolation=none \
  --test-concurrency=1 test/utils/tool-args.test.ts test/test-blank-filters.ts \
  test/test-list-issues.ts test/test-list-merge-requests.ts test/test-list-milestones.ts
npx prettier --check utils/tool-args.ts
```

`tsc` clean, 45/45 tests pass, `utils/tool-args.ts` formatting clean.

`test/utils/tool-args.test.ts` reports a pre-existing prettier warning at
lines 234-241 (`assert.deepEqual([...])` on an earlier test), untouched here.

## Relation to #753

#753 fixed blank entries at the query-building layer. This closes the same blank
values at the layer that decides the id/username conflict, which #753 could not see
because it runs before the query is built.


zereight:
@cursor please review this pull request.

Context: this is a follow-up to #753, which fixed blank entries inside array filters
at the query-building layer. This PR closes the same blank values one layer earlier, in
`cleanMutuallyExclusiveIdUsernameOptions`, which runs before the query is built.

What to check specifically:

1. `hasUsernameFilterValue` now uses `dropBlankArrayEntries` + `isBlankFilterValue`, so a
   blank username no longer deletes the id filter. Verify the id filter is still deleted
   in every case where a username actually reaches the query string, since that is the
   combination GitLab rejects with a 400.
2. The predicate's `undefined` / `null` early return — is it needed, or does
   `isBlankFilterValue` already cover those two and the branch is dead code?
3. Whether any of the three call sites (`list_issues`, `list_merge_requests`,
   `list_group_merge_requests`) needs its own test rather than only the shared helper.
4. `["", 0]` and similar mixed arrays: blank string entries are dropped, non-string
   entries are kept. Confirm that is the right reading of "blank".

Verification run locally: `npx tsc --noEmit` clean, 45/45 tests pass across
`test/utils/tool-args.test.ts`, `test/test-blank-filters.ts`, `test/test-list-issues.ts`,
`test/test-list-merge-requests.ts`, `test/test-list-milestones.ts`, and
`utils/tool-args.ts` is prettier-clean.

zereight:
@cursor Both of your suggestions are in `e3b30a0`.

**1. The scalar blank test — added.** You were right that this was the missing case, and right about why: `Boolean("  ")` was true, so a whitespace-only `assignee_username` deleted the id and then got dropped by the blank guard, leaving the request with no filter. The new case runs `""`, `" "`, and `"  \t "` and asserts the id survives:

```
node --test test/utils/tool-args.test.ts   # 20 tests, 20 pass, 0 fail
# with the pre-#763 predicate restored, it fails:
#   AssertionError: blank scalar " " must not drop the id
```

**2. `undefined` dropped from the guard — done.** You were right that it is unreachable: `isBlankFilterValue()` tests `value === undefined` itself, so `!isBlankFilterValue(dropBlankArrayEntries(undefined))` is already `false`. The `null` branch is the only one left, with a comment saying why it is not redundant:

```ts
function hasUsernameFilterValue(value: unknown): boolean {
  // `undefined` needs no branch: isBlankFilterValue() already counts it as blank. `null`
  // does not, and reporting it as a value would let the cleaner keep `assignee_id` while
  // the serializer writes `assignee_username=null`.
  if (value === null) {
    return false;
  }

  return !isBlankFilterValue(dropBlankArrayEntries(value));
}
```

**3. Skipped, per your call.** No call-site copies on `list_merge_requests` / `list_group_merge_requests` (their username fields are `z.string()`, so the array cases would not buy anything) and no HTTP-level `{ assignee_id: "7", assignee_username: ["", "  "] }` case. You marked both optional and I am not spending a round trip on them.

**4. The `null` mismatch stays documented, not fixed.** Widening `isBlankFilterValue` would change every list tool's query building, and `sanitizeToolArguments` (`index.ts:10842`) strips top-level `null` before the schemas parse, so the tool path cannot reach it. It is in the PR description under *Not in scope*.

`tsc` clean, prettier clean on the touched files, CI running on the new head. If the scalar case was the last thing you wanted before a clean verdict, say so and I will stop here.

zereight:
@cursor You were right about the polarity, and I fixed the comment — `8e182a2`, comment only, no behavior change.

Confirmed your reading against the runtime rather than by reasoning:

```
isBlankFilterValue(null) = false
appendFilterParam(new URLSearchParams(), "assignee_username", null) → "assignee_username=null"
```

So without the branch the cleaner **drops** `assignee_id` and the serializer writes `assignee_username=null` — the two agree that null is a value. With the branch the cleaner **keeps** the id while the serializer still writes the null, which is the id/username pair GitLab rejects. The branch selects that behavior, it does not prevent it. The comment now says exactly that, and names what would actually fix it (`isBlankFilterValue()` treating null as blank, in its own change, because it widens every list tool's query building).

The branch itself stays, for the reason we agreed on: it preserves `Boolean(null) === false`, and `sanitizeToolArguments` keeps top-level nulls out of the tool path.

You said this was not worth another round and the verdict was LGTM, so I am not asking you to re-review the whole PR. One line is enough: is the comment accurate now, and does `8e182a2` keep the LGTM?

```
npx tsc --noEmit                            # clean
npx prettier --check utils/tool-args.ts     # clean
node --test test/utils/tool-args.test.ts    # 20/20, 0 fail
```

